PR #1, opened 2017-02-09
Initial implementation

rslota:
This PR introduces... well, this project's implementation. This is, I would say, the first draft. It works and it has integration tests (100% coverage, and I hope it will stay this way). There are also some docs (mostly README.md) and several handy integrations (travis, credo, coveralls.io etc.). 

Even though there is `Dockerfile` on my TODO list, I think I'll skip it and we may add it later on. It's not needed, just very convenient. 

TODO:
- [x] tests (100% coverage :))
- [x] coveralls.io integration
- [x] dialyxir integration
- [x] credo integration (linter)
- [x] travis build integration (coverage + dialyzer + code linter)
- [x] Code docs / specs
- [x] Write README.md
- [ ] Dockerfile
- [x] Change SVG bages in README.md to point to master branch just before merging


config/prod.exs
```diff
@@ -0,0 +1,38 @@
+use Mix.Config
+
+config :maru, MongoosePush.Router,
+    versioning: [
+        using: :path
+    ],
+    https: [
+        ip: {127, 0, 0, 1},
```
michalwski, line 8: How about `{0, 0, 0, 0}` by default?

mix.exs
```diff
@@ -29,6 +42,19 @@ defmodule MongoosePush.Mixfile do
   #
   # Type "mix help deps" for more examples and options
   defp deps do
-    []
+    [
+     {:pigeon, git: "https://github.com/rslota/pigeon.git", tag: "6d1e4e3"},
```
michalwski, line 46: I see `rslota/pigeon` is a fork from `cstar/pigeon`. Are you @rslota going to send a PR to upstream?

priv/apns/dev_cert.pem
```diff
@@ -0,0 +1,23 @@
+-----BEGIN CERTIFICATE-----
```
michalwski: Can all these `*.pem` files be generated on demand? For example when the project is first compiled or run?

tools/travis-setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e
+
+# Skip this setup for jobs that don't run exunit
+test "${PRESET}" == "exunit" || exit 0
+
+docker run -d -p 2197:2197 mobify/apns-http2-mock-server
```
michalwski: Any chance for a specific version here and in the next container?

michalwski (commented):
This is simple, indeed :) I have some comments to this PR.

Also in a following PR I think we should add some metrics counting number of requests from clients, number of failed and succeeded requests to `FCM` and `APNS`. Elapsed time measurements for the whole request and specific requests to providers may prove useful on production as well.

config/prod.exs
rslota: I don't think its a good idea. In many cases this project will be used on same host as `MongooseIM`. Also if somebody uses it on different host, he will notice that he has to change configuration because REST API won't work without it. But with `{0, 0, 0, 0}` user may not notice that his service is open for attacks, because it will always work this way, even if such configuration is not needed.

mix.exs
rslota: I can try, but there are some changes that the author may not like. Author of this project had much different plan when comes to pooling.

tools/travis-setup.sh
rslota: We want the latest. Always. If the API changes we need to adapt, because our code needs to be compatible with newest APNS / FCM API.

priv/apns/dev_cert.pem
rslota: They can be generated but it adds one and only one additional step to building the project. If I leave those example certs in repo, there not need to have openssl while running the project and there not need for Makefile.

config/prod.exs
michalwski: Makes sense, thanks for the explanation.

mix.exs
michalwski: Try to convince him :)

priv/apns/dev_cert.pem
michalwski: I believe openssl is required in order to provide HTTPs from Elixir node. This probably means that openssl will have to be installed anyway. My biggest concern is with such files included in the repo, users may start using them without regenerating. For such a simple project which most probably we'll be used internally fresh certs and keys may not be mission critical. It is still sth which lights a warning in my brain, though.

tools/travis-setup.sh
michalwski: Fair enough :)

priv/apns/dev_cert.pem
rslota: Ok, you can now generate those certs via `mix certs.dev`.

README.md
```diff
@@ -1,19 +1,144 @@
...
 
+### Docker
+
+Soon :)
+
+### Local build
+
+#### Perquisites
+
+* Elixir (http://elixir-lang.org/install.html)
```
mentels: Which version of Elixir?

README.md
```diff
@@ -1,19 +1,144 @@
...
+After this step you may try to run the service via:
+```bash
+_build/prod/rel/mongoose_push/bin/mongoose_push console
+```
+
+Yeah, I know... It crashed. Running this service is fast and simple but unfortunately you can't have push notifications without properly configured `FCM` and/or `APNS` service. So, lets configure it!
+
+## Configuration
+
+The whole configuration is contained in file `config/{prod|dev|test}.exs` depending on with `MIX_ENV` you will be using. You should use `MIX_ENV=prod` for production installations and `MIX_ENV=dev` for you development. Anyway, lets take a look on `config/prod.exs`, part by part.
```
mentels: `on witch` -> `on which`

mentels: `for you development` -> ?

README.md
```diff
@@ -1,19 +1,144 @@
...
+
+Build step is really easy. Just type in root of the repository:
+```bash
+MIX_ENV=prod mix do certs.dev, deps.get, compile, release
+```
+
+After this step you may try to run the service via:
+```bash
+_build/prod/rel/mongoose_push/bin/mongoose_push console
+```
```
mentels, line 32: If we want to be serious, we *really need* to support releases. I won't accept an OS PR that has on support for releases :)

README.md
```diff
@@ -1,19 +1,144 @@
...
+        pool_size: 5,
+        mode: :prod
+    ]
+  ]
 ```
 
-Documentation can be generated with [ExDoc](https://github.com/elixir-lang/ex_doc)
-and published on [HexDocs](https://hexdocs.pm). Once published, the docs can
-be found at [https://hexdocs.pm/mongoose_push](https://hexdocs.pm/mongoose_push).
+The first level keyword list is a pool definition. You may have several named pools of different sizes and with different configurations. Currently the only reason you may want to do this is that, the `REST` client may switch between them by specifying matching `:mode` in their push request.
```
mentels: "The first level keyword list is a pool definition." - this is not very understandable. Maybe start with "to setup a pool...".

lib/mix/task_certs_dev.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Mix.Tasks.Certs.Dev do
+  use Mix.Task
+
+  @shortdoc "Generate fake certs (placeholders) for HTTPS endpoint and APNS"
+  @moduledoc """
```
mentels: @moduledoc is usually put at the very top.

lib/mix/task_certs_dev.ex
```diff
@@ -0,0 +1,56 @@
...
+                   "mongoose-push-apns-prod")
+  end
+
+  defp maybe_gen_https do
+    maybe_gen_cert("priv/ssl/fake_cert.pem", "priv/ssl/fake_key.pem",
+                   "mongoose-push")
+  end
+
+  defp maybe_gen_cert(cert_file, key_file, common_name) do
+    unless File.exists?(cert_file) and File.exists?(key_file) do
```
mentels: This condition is not readable. If I need to think what it means it has to be structured differently. One option would be:

```elixir
case {File.exists?(cert_file), File.exists?(key_file)} do
    {false, false} -> ...
    _ -> ...
end
```

lib/mongoose_push.ex
```diff
@@ -1,18 +1,106 @@
...
+  for both services.
+
+  At this moment only those two services are supported but in future
+  MongoosePush may and probably will support even more Push Notification Services.
   """
 
-  @doc """
-  Hello world.
+  require Logger
+  import MongoosePush.Application
```
mentels: No, I really discourage these import. Because of that it takes you time to find functions like `pools_by_name`

lib/mongoose_push.ex
```diff
@@ -1,18 +1,106 @@
...
+  `:mode` option is also specific to APNS but it only selects appropriate
+  worker pool (with `:mode` set to either `:prod` or `:dev`).
+  Default value to `:mode` is `:prod`.
   """
-  def hello do
-    :world
+  @spec push(String.t, request) :: :ok | {:error, term}
+  def push(device_id, %{:service => service} = request) do
+      mode = Map.get(request, :mode, :prod)
+      [pool | _] = pools_by_mode(service, mode)
```
mentels: It looks like the functions on lines 44 and 45 are consructing a worker name. For me this level of abstraction doesn't belong to this function.

lib/mongoose_push.ex
```diff
@@ -1,18 +1,106 @@
...
+    gcm_notification = Pigeon.GCM.Notification.new(device_id, msg)
+
+    gcm_notification
+    |> GCM.push([name: worker])
+    |> normalize_response(:fcm, device_id)
   end
+
+  @spec push(service, atom | pid, String.t, request) :: :ok | {:error, term}
+  defp push(:apns, worker, device_id, request) do
+    raw_notification = prepare_notification(:apns, request)
```
mentels: This type of code can be rewritten as:

```elixir
raw_notification = 
     prepare_notification(:apns, request)
     |> APNS.Notification.new(device_id, request[:topic])
     |> APNS.push([name: worker])
     |> normalize_response(:apns, device_id)
```

lib/mongoose_push.ex
```diff
@@ -1,18 +1,106 @@
...
+    %{
+      "alert" => %{
+        "title" => request.title,
+        "body" => request.body
+      },
+      "badge" => request[:badge],
+      "category" => request[:click_action]
+    }
+  end
+
```
mentels: For me, this module should not distinguish between the two types of the backends further than :fcm and :apns. In other words, I believe that the two deserve their own modules like `MongoosePush.FCM` and `MongoosePush.APNS` and the specifics of these services should sit in them. Maybe the next PR would address that but I simply don't like it.

mentels (commented):
Will continue the review from lib/mongoose_push/api/v1.ex tomorrow.

lib/mongoose_push.ex
rslota: Yeah, I did write it this way, but `credo` does not agree with you. Its says that each pipe operator has to start with variable, not expression.

lib/mongoose_push.ex
rslota: Thats true, `alias` seems to be a better idea.

README.md
rslota: Could you elaborate? What do you mean by `support for releases` exactly? This `README.md` shows how to run this project from the release.

README.md
rslota: your* :)

lib/mix/task_certs_dev.ex
rslota: I think that readability is very subjective, especially in this case (huh, `case` :)). This `case` pattern may be easier to read for `erlang` devs since in `erlang` this is the only sane way to write such condition. Elixir makes it easier and in my opinion more readable, so I wanted to exploit that.  

But anyway, this is so minor that I'm just gonna change it, I don't really like `unless` that much :)

lib/mongoose_push/api/v1.ex
```diff
@@ -0,0 +1,41 @@
...
+        device_id = params.device_id
+        case MongoosePush.push(device_id, Map.delete(params, :device_id)) do
+          :ok ->
+            conn
+            |> put_status(200)
+            |> json(nil)
+          {:error, reason} ->
+            conn
+            |> put_status(500)
+            |> json(%{:details => inspect reason})
```
mentels: The module is beautiful. Simple and readable. One issue though, the `inspect` from the `Kernel` module is not to be used to format data that you return to the client. You have to format you data differently, without inspect.

lib/mongoose_push/application.ex
```diff
@@ -4,19 +4,126 @@ defmodule MongoosePush.Application do
...
+
+    config
+    |> Enum.map(&(elem(&1, 0)))
+  end
+
+  def pools_by_mode(:apns = service, mode) do
+    config = env(service)
+
+    config
+    |> Enum.group_by(&(mode(elem(&1, 1))), &(elem &1, 0))
```
mentels: This is not really readable.

lib/mongoose_push/application.ex
```diff
@@ -4,19 +4,126 @@ defmodule MongoosePush.Application do
...
+      pool_config = translate_worker_config(:fcm, pool_config)
+
+      Enum.map(1..pool_size(:fcm, pool_name), fn(id) ->
+        worker_name = worker_name(:fcm, pool_name, id)
+        worker(Pigeon.GCMWorker, [worker_name, pool_config], [id: worker_name])
+      end)
+    end)
+
+    workers
+    |> List.flatten
```
mentels: Instead of flattening the list afterwards, you could use `Enum.flat_map/2`.

lib/mongoose_push/router.ex
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  rescue_from :all, as: e do
+    status = Map.get e, :plug_status, 500
+    log_level =
+        case status >= 500 do
+          true  -> :error
+          false -> :info
+        end
+    Logger.log log_level, inspect e
```
mentels, line 58: Again, you shouldn't be using inspect to format your output data.

lib/mongoose_push/service.ex
```diff
@@ -0,0 +1,9 @@
+defmodule MongoosePush.Service do
```
mentels, line 1: Cool :)

README.md
mentels: Damn...how could I have missed that? Sorry, of course you do support releases with distillery. Awesome :)

test/mongoose_push_application_test.exs
```diff
@@ -0,0 +1,63 @@
...
+  doctest MongoosePush.Application
+
+  test "worker name" do
+    assert :apns_name1_1 == worker_name(:apns, :name1, 1)
+    assert :fcm_name2_12 == worker_name(:fcm, :name2, 12)
+  end
+
+  test "pool groups" do
+    assert [:dev1, :dev2] == pools_by_mode(:apns, :dev)
+    assert [:prod1, :prod2] == pools_by_mode(:apns, :prod)
```
mentels: How do I known that `[:prod1, :prod2]` is the expected value? Why two ?

test/mongoose_push_application_test.exs
```diff
@@ -0,0 +1,63 @@
...
+    assert [:default] == pools_by_mode(:fcm, :default)
+  end
+
+  test "pool size" do
+    assert 1 == pool_size(:apns, :dev1)
```
mentels: Same here, where the correct values come from?

test/mongoose_push_application_test.exs
mentels: Ahh I know, there's the `config/test.exs`. Hmm if someone changes this file the test won't pass and assertion error will be printed. Seems to me that then the error becomes not obvious. One way of tackling this would be to have simple assertion like `Application.get_env(:mongoose_push, :sth)[:pools] == 4` (that's pseudo code ofc)

mentels: Or you could put these assertions agains the test file in the setup. They wouldn't be event assertion, rather pattern matchings so that we can clearly show that the tests relay on some particular config.

test/mongoose_push_test.exs
```diff
@@ -1,8 +1,152 @@
...
+        :click_action => "click.action",
+        :tag => "tag value"
+      }
+    fail_tokens(:fcm, [%{device_token: "androidtestdeviceid65", status: 200,
+                         reason: "InvalidRegistration"}])
+
+    assert {:error, _} = push("androidtestdeviceid65", notification)
+  end
+
+  defp reset(service) do
```
mentels: hmm, looks like the second clause of this function, which starts on line 107, has the same body.

test/rest_v1_test.exs
```diff
@@ -0,0 +1,75 @@
...
+     %Response{status_code: status_code} =
+       HTTPoison.post!("https://localhost:8443" <> path, Poison.encode!(json),
+                       [{"Content-Type", "application/json"}],
+                       hackney: [:insecure])
+     status_code
+  end
+
+  defp get(path) do
+    %Response{status_code: status_code} =
+      HTTPoison.get!("https://localhost:8443" <> path, [], hackney: [:insecure])
```
mentels, line 71: Why are you using the `insecure` hackney option?

lib/mix/task_certs_dev.ex
mentels: Maybe you can simply write `if File.exists? ... and ... do ok else do_stuff() end`

lib/mongoose_push.ex
```diff
@@ -1,18 +1,52 @@
...
+  for both services.
+
+  At this moment only those two services are supported but in future
+  MongoosePush may and probably will support even more Push Notification Services.
   """
 
-  @doc """
-  Hello world.
+  require Logger
+  alias MongoosePush.Application
```
mentels: I don't like this alias. It's misleading as at first glance you don't know if it refers to `Elixir.Application` or to `Elixir.MngoosePush.Application`.

lib/mongoose_push/application.ex
```diff
@@ -4,19 +4,126 @@ defmodule MongoosePush.Application do
   @moduledoc false
```
mentels, line 4: For me, the whole pools management doesn't belong to this module. If would prefer just having it separated out to sth like MongoosePush.Pools which would provide some API to get workers names to start then under supervisors and then the other one to select a woker. Even better, I believe, is a solution in which you have a separate supervisors for each type of pool and it knows more details and the the "pools" module that know even more details.

lib/mongoose_push/application.ex
```diff
@@ -4,19 +4,126 @@ defmodule MongoosePush.Application do
...
+  @spec env(atom) :: term
+  def env(var), do: Application.get_env(:mongoose_push, var)
+
+  defp fcm_workers(nil), do: []
+  defp fcm_workers(config) do
+    workers = Enum.map(config, fn({pool_name, pool_config}) ->
+      pool_config = translate_worker_config(:fcm, pool_config)
+
+      Enum.map(1..pool_size(:fcm, pool_name), fn(id) ->
+        worker_name = worker_name(:fcm, pool_name, id)
```
mentels: To deep... doesn't credo complain about that?

test/rest_v1_test.exs
rslota: Because this is connection to `HTTPS` endpoint that is running on self-signed certs generated for tests. Without this option connection would fail due to `unknown_ca` error.

lib/mongoose_push.ex
rslota: Ok, you're right.

test/mongoose_push_test.exs
rslota: Yes, this is clearly simple duplication.

test/rest_v1_test.exs
mentels: Thanks

test/mongoose_push_application_test.exs
rslota: I've put some assertions in setup, but I don't check every single thing in the config. There are only some general pools checks to show that they are needed.

lib/mongoose_push/application.ex
rslota: No it does not. This is only 2 level in, so I guess its maximum allowed.

lib/mongoose_push/application.ex
rslota: Pools have been refactored. Please review new implementation :)

lib/mongoose_push/api/v1.ex
rslota: Removed.

lib/mongoose_push/pools.ex
```diff
@@ -0,0 +1,44 @@
+defmodule MongoosePush.Pools do
```
mentels, line 1: Now this is awesome

test/mongoose_push_application_test.exs
```diff
@@ -1,25 +1,23 @@
...
   doctest MongoosePush.Application
 
-  test "worker name" do
-    assert :apns_name1_1 == worker_name(:apns, :name1, 1)
-    assert :fcm_name2_12 == worker_name(:fcm, :name2, 12)
-  end
+  setup do
+    # Validate config/text.exs that is need for this test suite
+    apns_pools = Keyword.keys(Application.get_env(:mongoose_push, :apns))
+    4 = length(apns_pools)
```
mentels: That could be reduced to [:prod1, :prod2, :dev1, :dev2] = apns_pools

But is so minor that you don't have to bother yourself.

mentels (approved):

PR #7, opened 2017-03-06
Feature/metricks

rslota:
Add metrics support to MongoosePush. For now this PR is only for build on travis.


mix.exs
```diff
@@ -24,15 +24,27 @@ defmodule MongoosePush.Mixfile do
...
+
+     # Just overrides to make elixometer compile...
+     {:setup, github: "uwiger/setup", tag: "1.8.0", override: true, manager: :rebar},
+     {:edown, github: "uwiger/edown", tag: "0.8", override: true},
+     {:lager, ">= 3.2.1", override: true},
+     {:exometer_core, github: "PSPDFKit-labs/exometer_core", override: true},
+     {:exometer, github: "PSPDFKit-labs/exometer"},
+     {:elixometer, github: "pinterest/elixometer"},
+
+
```
mongoosepushbot, line 44: There should be no more than 1 consecutive blank lines.

lib/mongoose_push/metrics.ex
```diff
@@ -0,0 +1,24 @@
+defmodule MongoosePush.Metrics do
+  use Elixometer
+
+  def update(return_value, metric_prefix, value \\ 1) do
```
mongoosepushbot, line 4: Functions should have a @spec type specification.

lib/mongoose_push/metrics.ex
```diff
@@ -0,0 +1,24 @@
+defmodule MongoosePush.Metrics do
```
mongoosepushbot, line 1: Modules should have a @moduledoc tag.

lib/mongoose_push.ex
```diff
@@ -36,13 +37,26 @@ defmodule MongoosePush do
...
       worker = Pools.select_worker(service, mode)
       module = MongoosePush.Application.services()[service]
 
       notification = module.prepare_notification(device_id, request)
-      module.push(notification, device_id, worker)
+      opts = [timeout: 60_000]
+      push_result = module.push(notification, device_id, worker, opts)
+
+      push_result
+      |> MongoosePush.Metrics.update(~s"push.#{service}.#{mode}")
```
mongoosepushbot, line 52: Nested modules could be aliased at the top of the invoking module.

PR #8, opened 2017-03-08
Elixometer integration

rslota:
This PR introduces `elixometer` integration in order to support metrics. There is a `MongoosePush.Metrics` helper module for all functions that help with metrics management. For now theres only one function that increments a spiral/counter metric based on return value of the function. 

Also theres one new option in `:prod` configuration -  PUSH_HTTPS_ACCEPTORS - defines how many ranch acceptors have to be stared for `HTTP` endpoint of the MongoosePush application.

Depends on the #9 .


mix.exs
```diff
@@ -24,15 +24,26 @@ defmodule MongoosePush.Mixfile do
...
+
+     # Just overrides to make elixometer compile...
+     {:setup, github: "uwiger/setup", tag: "1.8.0", override: true, manager: :rebar},
+     {:edown, github: "uwiger/edown", tag: "0.8", override: true},
+     {:lager, ">= 3.2.1", override: true},
+     {:exometer_core, github: "PSPDFKit-labs/exometer_core", override: true},
+     {:exometer, github: "PSPDFKit-labs/exometer"},
+     {:elixometer, github: "pinterest/elixometer"},
+
+
```
mongoosepushbot: There should be no more than 1 consecutive blank lines.

lib/mongoose_push/metrics.ex
```diff
@@ -0,0 +1,24 @@
+defmodule MongoosePush.Metrics do
+  use Elixometer
+
+  def update(return_value, metric_prefix, value \\ 1) do
```
mongoosepushbot: Functions should have a @spec type specification.

lib/mongoose_push/metrics.ex
```diff
@@ -0,0 +1,24 @@
+defmodule MongoosePush.Metrics do
```
mongoosepushbot, line 1: Modules should have a @moduledoc tag.

lib/mongoose_push.ex
```diff
@@ -36,13 +37,26 @@ defmodule MongoosePush do
...
       worker = Pools.select_worker(service, mode)
       module = MongoosePush.Application.services()[service]
 
       notification = module.prepare_notification(device_id, request)
-      module.push(notification, device_id, worker)
+      opts = [timeout: 60_000]
+      push_result = module.push(notification, device_id, worker, opts)
+
+      push_result
+      |> MongoosePush.Metrics.update(~s"push.#{service}.#{mode}")
```
mongoosepushbot: Nested modules could be aliased at the top of the invoking module.

mongoosepushbot (approved):

mongoosepushbot (approved):

mongoosepushbot (approved):

mongoosepushbot (approved):

README.md
```diff
@@ -61,6 +61,7 @@ Environmental variables to configure production release:
 * `PUSH_HTTPS_PORT` - The port of the MongoosePush REST endpoint. Please not that docker exposes only `8443` port, so changing this setting is not recommended
 * `PUSH_HTTPS_KEYFILE` - Path to PEM keyfile used for REST endpoint
 * `PUSH_HTTPS_CERTFILE` - Path to PEM certfile used for REST endpoint
+* `PUSH_HTTPS_ACCEPTORS` - Number of TCP acceptors to start
```
michalwski, line 64: How is this option connected to `Elixometer`?

lib/mongoose_push/metrics.ex
```diff
@@ -0,0 +1,35 @@
...
+      {:error, _reason} ->
+        increment(metric_prefix <> ".error.all", value)
+        increment(metric_prefix <> ".error.unknown", value)
+    end
+    return_value
+  end
+
+  defp increment(metric, value) do
+    update_counter(metric, value)
+    update_spiral(metric, value)
```
michalwski: Why do we need counter and spiral? In `exometer` spiral metric provides total counter as well, could this be reused and number of metrics reduced twice?

mix.exs
```diff
@@ -24,20 +24,31 @@ defmodule MongoosePush.Mixfile do
...
      {:maru_swagger, github: "elixir-maru/maru_swagger"},
      {:distillery, "~> 1.0"},
      {:confex, "~> 1.4", override: true},
      {:mix_docker, "~> 0.3"},
+
+     # Just overrides to make elixometer compile...
+     {:setup, github: "uwiger/setup", tag: "1.8.0", override: true, manager: :rebar},
+     {:edown, github: "uwiger/edown", tag: "0.8", override: true},
+     {:lager, ">= 3.2.1", override: true},
+     {:exometer_core, github: "PSPDFKit-labs/exometer_core", override: true},
```
michalwski, line 40: Could you explain why `PSPDFKit-labs/exmometer*` are used instead of the original ones form `Feuerlabs`?

michalwski (commented):
Thanks for the metrics! I left some comments in the code also, the PR lacks a description. I think it's a good practice to give a bit longer description than just a title. Especially when the PR does a bit more then what's stated in the title.

Also is it possible to use `exometer` like metric names (list, ie [push, failures]) instead for strings?

README.md
rslota: It's not, I forgot about that. I can move it to separate PR if you want to. Theres no good place where it fits.

michalwski: It can be in the same PR as long as there is some note about it in the PR desc.

mix.exs
rslota: This is a version with stripped down dependancies. The original one uses lots of deps that conflict with `exometer`. It's not pretty, but the author of exometer made a lot of mess while splitting `exometer` into two repos. I've tried several approaches to make it work and this seems to be the simples one. Also this was suggested by author of the `elixometer`

lib/mongoose_push/metrics.ex
rslota: Well, thats "to discuss" I guess. I'm not sure how those will be used, they may have different purpose and may be reseted by other conditions. 

To be perfectly honest I've used both of them because I wanted to make sure that `elixometer` fully works (knowing the problems I had to setup in the projects, I wasn't sure). 
I know that having lots of metrics may be hard on the system, but here, we have one.

To sum up: One word and I'll remove it.

lib/mongoose_push/metrics.ex
rslota: But it has :<

README.md
rslota: Oh, I don't know how this happened. I must have simple forgot description after reorganising current PRs. Well, there is one now :)

rslota:
> Also is it possible to use exometer like metric names (list, ie [push, failures]) instead for strings?

No it's not. This the the syntax that `elixometer` expects, because the first things it does is splitting the string on dot ('.') to pass it to `exometer`.

lib/mongoose_push/metrics.ex
michalwski: I pretty sure it's enough to have only spiral here which gives you total count and count in last minute. Metric of type counter could be useful for things which needs to be decremented also. Here, to my understanding we are only incrementing.

michalwski:
> because the first things it does is splitting the string on dot ('.') to pass it to `exometer`.

So clever `elixometer`! I guess there is nothing we can do here instead of wasting some resources on this unnecessary conversion.

mongoosepushbot (approved):

michalwski:
Thanks!

PR #10, opened 2017-03-09
Speed up metrics

rslota:
Replace functions of `MongoosePush.Metrics` module with macros to speed up construction of metric name (compile time vs runtime).


mongoosepushbot (approved):

michalwski:
Thanks and sorry for the delay.

PR #17, opened 2017-04-13
Support data payload

rslota:
This PR adds support for custom data payload for both APNS and FCM.

### DIFF
--- a/lib/mongoose_push/api/v1.ex
+++ b/lib/mongoose_push/api/v1.ex
@@ -17,6 +17,7 @@ defmodule MongoosePush.API.V1 do
     optional  :click_action,  type: String
     optional  :tag,           type: String
     optional  :topic,         type: String
+    optional  :data,          type: &(&1) # Use raw json value to skip all maru's validators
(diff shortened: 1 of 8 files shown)


lib/mongoose_push/api/v1.ex
```diff
@@ -17,6 +17,7 @@ defmodule MongoosePush.API.V1 do
     optional  :click_action,  type: String
     optional  :tag,           type: String
     optional  :topic,         type: String
+    optional  :data,          type: &(&1) # Use raw json value to skip all maru's validators
```
mongoosepushbot: Line is too long (max is 80, was 92).

michalwski (commented):
Could you fix the too long line issue?

michalwski:
Also sth is wrong with tests on Travis

rslota:
> Also sth is wrong with tests on Travis

That's soo weird... I noticed that before with `jerboa` repo. There also somehow commits from non-master branch couldn't be fetched but only on travis.

Fixed.

PR #18, opened 2017-04-14
Support silent notifications

rslota:
This PR introduces new API for "silent" push notifications. The new type of notifications is possible to generate only with new REST `API v2` - making in possible in `API v1` would introduce to much "code smell".


lib/mongoose_push/service/fcm.ex
```diff
@@ -11,12 +11,18 @@ defmodule MongoosePush.Service.FCM do
 
   @spec prepare_notification(String.t(), MongoosePush.request) ::
     Service.notification
+  def prepare_notification(device_id, request = %{alert: nil}) do
```
mongoosepushbot: File has the variable name before the pattern while most of the files have the variable name after the pattern when naming parameter pattern matches

lib/mongoose_push/service/apns.ex
```diff
@@ -12,17 +12,23 @@ defmodule MongoosePush.Service.APNS do
 
   @spec prepare_notification(String.t(), MongoosePush.request) ::
     Service.notification
+  def prepare_notification(device_id, request = %{alert: nil}) do
```
mongoosepushbot: File has the variable name before the pattern while most of the files have the variable name after the pattern when naming parameter pattern matches

lib/mongoose_push/api/v2.ex
```diff
@@ -0,0 +1,52 @@
...
+
+  plug Plug.Parsers,
+      pass: ["application/json", "text/json"],
+      json_decoder: Poison,
+      parsers: [:urlencoded, :json, :multipart]
+
+  params do
+    requires  :service,       type: Atom, values: [:fcm, :apns]
+    optional  :mode,          type: Atom, values: [:prod, :dev]
+    optional  :topic,         type: String # Only for APNS, alert/data independent
```
mongoosepushbot: Line is too long (max is 80, was 82).

lib/mongoose_push/api/v2.ex
```diff
@@ -0,0 +1,52 @@
...
+
+    optional  :alert,         type: Map do
+      requires  :body,          type: String
+      requires  :title,         type: String
+      optional  :badge,         type: Integer
+      optional  :click_action,  type: String
+      optional  :tag,           type: String
+    end
+
+    optional  :data,          type: &(&1) # Use raw json value to skip all maru's validators
```
mongoosepushbot: Line is too long (max is 80, was 92).

lib/mongoose_push/api/v2.ex
```diff
@@ -0,0 +1,52 @@
...
+    at_least_one_of [:alert, :data] # We need to send at least one of data or alert
```
mongoosepushbot: Line is too long (max is 80, was 83).

lib/mongoose_push/api/v1.ex
```diff
@@ -17,14 +17,21 @@ defmodule MongoosePush.API.V1 do
     optional  :click_action,  type: String
     optional  :tag,           type: String
     optional  :topic,         type: String
+    optional  :data,          type: &(&1) # Use raw json value to skip all maru's validators
```
mongoosepushbot: Line is too long (max is 80, was 92).

lib/mongoose_push.ex
```diff
@@ -45,6 +52,11 @@ defmodule MongoosePush do
       worker = Pools.select_worker(service, mode)
       module = MongoosePush.Application.services()[service]
 
+      request = # Just make sure both data and alert keys exist for convenience (by may be nil)
```
mongoosepushbot: Line is too long (max is 80, was 95).

michalwski (approved):
Looks good to me. One question, how much sense does it make to keep the old API version?

rslota:
> Looks good to me. One question, how much sense does it make to keep the old API version?

Initially I just wanted to test and showcase that it indeed works as expected, but now that I know that we have users that already use this project we really should keep the `v1` API.

rslota:
@michalwski Since the `0.1.0` version has been released and there are not comments to the code, I think it's ready to merge. There is one more thing I'm working on regarding APNS `topics`, but that will be in another PR.

michalwski:
Thanks!

PR #24, opened 2017-04-19
Default APNS topic

### LINKED ISSUE #23: Default value for APNS Topic
The APNS topic in production environments have to be set via MongoosePush API. This may be redundant since, MongoosePush may default this topic to the first one available in APNS X509 certificate.

rslota:
This PR resolves #23 .
The APNS topic is extracted from the APNS certificate and it's used as default value if the topic is not set via the MongoosePush API on per-notification basis. 

The PR is not ready yet. The feature is complete but there are no tests and there are issues with `maru` conflicting with erlang sources. 

TODO:
- [x] Fix tests compile errors related to `maru` not tolerating erlang modules 
- [x] Add tests for this feature


michalwski:
Did the coverage really decreased to 53% from 100%? Or is sth wrong with coveralls?

rslota:
> Did the coverage really decreased to 53% from 100%? Or is sth wrong with coveralls?

There's a new erlang source file that is generated and used in only like ~30%, so the code coverage dropped. I will exclude this file from the coverage report.

Or even better - maybe it's good idea to just remove this file because it is recompiled every time anyways.

src/APNS-Topics.erl
```diff
@@ -0,0 +1,268 @@
+%% Generated by the Erlang ASN.1 BER_V2-compiler version, utilizing bit-syntax:4.0.4
+%% Purpose: encoder and decoder to the types in mod APNS-Topics
+
+-module('APNS-Topics').
```
michalwski: Do we need this file in the repo? As far as I understand it's generated before compilation.

rslota: I've removed this file.

README.md
```diff
@@ -79,13 +79,14 @@ Environmental variables to configure production release:
...
 
 ##### Settings for production APNS service:
 * `PUSH_APNS_PROD_ENDPOINT` - Hostname of `APNS` service. Set only for local testing. By default this option points to the Apple's official hostname
 * `PUSH_APNS_PROD_CERT` - Path Apple's production certfile used to communicate with `APNS`
 * `PUSH_APNS_PROD_KEY` - Path Apple's production keyfile used to communicate with `APNS`
 * `PUSH_APNS_PROD_USE_2197` - `true`/`false` - Enable or disable use of alternative `2197` port for `APNS` connections in production mode. Disabled by default
-* `PUSH_APNS_PROD_POOL_SIZE` - Connection pool size for `APNS` service in production mode
+* `PUSH_APNS_PROD_POOL_SIZE` - Connection pool size for `APNS` service in production modeDefault `APNS` topic to be set if the client app doesn't specify it with the API call. If this option is not set, MongoosePush will try to extract this value from the provided APNS certificate (the first topic will be assumed default)Connection pool size for `APNS` service in production mode
```
mentels: Looks like some part of the doc was accidentally pasted twice?

lib/mix/tasks/certs_dev.ex
```diff
@@ -0,0 +1,106 @@
...
+  `mobify/apns-http2-mock-server`).
+  """
+  @shortdoc "Generate fake certs (placeholders) for HTTPS endpoint and APNS"
+
+  use Mix.Task
+
+  # From: https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html
+  @apns_topic_extn_id {1, 2, 840, 113635, 100, 6, 3, 6}
+  # Here we use the binary extension extracted from real APNS certificate. It's much better for
+  # testing to use real extension instead of generated one since, the genertor would be based
```
mentels: `a real one` `the generated one since the` 

Probably @Dzol can confirm that.

lib/mix/tasks/certs_dev.ex
```diff
@@ -0,0 +1,106 @@
...
+    cert_dir = Path.dirname(cert_file)
+    key_dir = Path.dirname(key_file)
+
+    :ok = File.mkdir_p(cert_dir)
+    :ok = File.mkdir_p(key_dir)
+
+    ext_file = openssl_tmp_extfile(extensions)
+    req_file = cert_file <> ".csr"
+
+    # Create CSR
```
mentels: Instead of having the comments extracted the related code blocks into functions.

lib/mix/tasks/certs_dev.ex
```diff
@@ -0,0 +1,106 @@
...
+      "x509", "-req", "-days", "365", "-in", req_file, "-signkey", key_file,
+      "-extfile", ext_file, "-out", cert_file
+    ])
+
+    :ok = File.rm!(ext_file)
+    :ok = File.rm!(req_file)
+  end
+
+  defp openssl_tmp_extfile(extensions) do
+    ext_file = "/tmp/" <> UUID.uuid4()
```
mentels: Use `Path.join` to construct a path.

lib/mix/tasks/compile_asn1.ex
```diff
@@ -0,0 +1,38 @@
...
+  use Mix.Task
+
+  @asn1_src "asn.1"
+  @erl_src "src"
+
+  @spec run(term) :: :ok
+  def run(_) do
+    :ok = File.mkdir_p!(@erl_src)
+    for file <- ls_r!(@asn1_src) do
+      case Path.extname(file) do
```
mentels: Too deep for me. `case` in `for`

lib/mongoose_push/pools.ex
```diff
@@ -38,7 +38,7 @@ defmodule MongoosePush.Pools do
   @doc "Return random worker name for given service and with given `:mode` set"
   @spec select_worker(MongoosePush.service, MongoosePush.mode) :: atom
   def select_worker(service, mode) do
-    [pool | _] = pools_by_mode(service, mode)
+    [pool | _] = __MODULE__.pools_by_mode(service, mode)
```
mentels, line 41: Why this change?

lib/mongoose_push/service/apns/certificate.ex
```diff
@@ -0,0 +1,46 @@
...
+  def extract_topics!(cert_file) do
+    maybe_extension =
+      cert_file
+      |> File.read!()
+      |> :public_key.pem_decode()
+      |> List.keyfind(:Certificate, 0)
+      |> get_cert_extension!(@apns_topic_extn_id)
+
+    if maybe_extension == nil do
+        throw :no_extension
```
mentels, line 24: If `maybe_extension == nil` is an error condition the use `raise`.

test/mongoose_push_test.exs
```diff
@@ -1,14 +1,84 @@
...
+      ]) do
+        assert :ok == push(@test_token, %{:service => :apns, :title => "", :body => "", mode: :dev})
+
+        apns_request = last_activity(:apns)
+        assert @test_token == apns_request["device_token"]
+        assert "dev_topic" == apns_request["request_headers"]["apns-topic"]
+      end
+    end
+
+    test "is not set if theres no default valie in config nor certificate" do
```
mentels: `value`? no `valie`?

mentels (changes requested):

lib/mix/tasks/certs_dev.ex
Dzol: The _the_?

lib/mongoose_push/service/apns/certificate.ex
rslota: Well, the issue here is that this whole code can crash due to... well, reasons. There is one catch for that, only in order to log the issue properly. Here I throw `:no_extension` only because badmatch on `nil` is not saying anything useful. This is not an error condition that changes logic somehow, at least not in different way then any other `error` here. This is just to make this one error more readable in logs.

README.md
rslota: Thanks, fixed.

lib/mix/tasks/certs_dev.ex
rslota: I would say that it should be exactly the opposite of that :)

lib/mix/tasks/certs_dev.ex
rslota: Good point, done.

lib/mix/tasks/certs_dev.ex
rslota: Done, I wonder how it happened...

lib/mix/tasks/compile_asn1.ex
rslota: I would argue with that, but luckily I don't have to, since I've just removed this case along with `ls_r!/1` anyway.

lib/mongoose_push/pools.ex
rslota: In order to enable mocking of this function.

test/mongoose_push_test.exs
rslota: Thanks!

rslota:
@mentels I've pushed some code quality changes you asked for.

lib/mongoose_push/pools.ex
mentels: How it works? `__MODULE__` is set at compile time... I don't understand.

lib/mongoose_push/service/apns/certificate.ex
mentels: An "elixir error" (raise "sth") won't be catched by your `try do catch` in `maybe_setup_default_apns_topic`. It will only catch the `throw`. If it's not different than any other error then it should case a whole thing to fail - as any other error would do.

I still see inconsistency here.

lib/mongoose_push/pools.ex
rslota: The point here is to call the function explicitly from `__MODULE__` module, because when you load mock, the original module name changes. Since `__MODULE__` is set in compile time, it still points to original name, which is now the mock module. If you would call function "locally", you would call the function without even touching the mock, which is not desirable.

lib/mongoose_push/service/apns/certificate.ex
rslota: This function is meant to `try` load APNS topic. If it fails, for *whatever* reason, `maybe_setup_default_apns_topic` should fallback to other means of getting the topic. Again - the reason of the failure could be less important. The only thing that the reason is helpful of is that we should log the failures in order to inform user that we have to fallback to other means of getting topic and possibly why. Therefore, this `if` is here only to change log message to be more readable in this one case. 

Elixir's `raise` is meant for special error conditions, that maybe need special handling. Here, theres no different route of handling this exception no matter if it looks like that, or is just a badmatch, which would be the case if the `if` wasn't there.

mentels (approved):

PR #33, opened 2017-08-06
Use mainstream version of Maru

rslota:


Dzol (approved):
It seems fine to me... what's up with this test mode though?

rslota:
> It seems fine to me... what's up with this test mode though?

It takes over, ignoring Maru Router's related settings in `config/test.exs`. Maybe in "test mode" testing is easier if this is new project, but since I have already tailored testing settings/methods on my own, I don't really see the point of adapting to the "new Maru's way" of doing that. And even is this would be in some way better, we can do that... well, later, for now I just wanted to switch repositories / upgrade deps.

deadjdona:
==> maru_swagger
Compiling 6 files (.ex)
warning: function Mix.Phoenix.base/0 is undefined (module Mix.Phoenix is not available)
  lib/maru_swagger/config_struct.ex:33

Generated maru_swagger app

PR #49, opened 2017-10-29
Added "mutable-content" attribute in the payload of APNS

Mo33n:
IOS 10 have handling for a new payload attribute,

By using  "mutable-content" notification service extensions can be written.


rslota (changes requested):
Thanks for the PR! 

Looks good to me, but I think we need to extend it a bit. In my opinion, this should be an option in MongoosePush's API. This type of alerts will always wake up the device, hence may not always preferable. Since this option probably should be disabled by default, we also need to add the switch to enable it in `MongooseIM` via client configuration (`enable` stanza option).  If you don't really have time for such change, I'll do it eventually, but it's gonna take some time for me until I get to it.

rslota:
This PR has been reworked with tests and API changes in #62 .

PR #52, opened 2017-11-14
Fix little grammar mistake

kanes115:


rslota (approved):
Thanks for the fixes!

PR #53, opened 2017-11-17
Helper openssl commands to conver pkcs12 files

michalwski:
I updated the README with `openssl` commands to get cert and key form pkcs12 files.


rslota:
Great, thanks!

PR #54, opened 2017-12-07
Fixup dizalyzer job and reduce number of docker image pushes on Travis

rslota:
There are two small changes:
* Name of dialyzer "preset" has been fixed - for some time now, because of this, dialyzer job hasn't been starting correctly. 
* Deduplication of docker image builds / pushes using Travis `stages` (Test -> Deploy)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,10 +27,12 @@ otp_release:
 env:
     - PRESET=exunit MIX_ENV=test
 
+script: tools/travis-test.sh
+
 matrix:
     include:
         - otp_release: 20.0
-          env: PRESET=dialyzer_only
+          env: PRESET=dialyzer MIX_ENV=test
 
 script: tools/travis-test.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,10 @@ script: tools/travis-test.sh
 
 matrix:
     include:
-        - otp_release: 20.0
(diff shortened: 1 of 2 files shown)


kzemek (approved):

michalwski (approved):
The stage approach looks good. Thanks!

.travis.yml
```diff
@@ -31,10 +31,10 @@ script: tools/travis-test.sh
...
+        - stage: test
+          otp_release: 20.0
           env: PRESET=dialyzer MIX_ENV=test
-
-script: tools/travis-test.sh
-
-after_success:
-  - tools/travis-docker.sh
+        - stage: deploy
+          elixir: 1.5.1
```
deadjdona: what's the reason to use 1.5.1 version?

most users will install latest version:
elixir/unknown,now 1.5.2-1 all [installed]

rslota: Well, technically Elixir is used here only in order to run mix. Deploy stage builds docker image which uses internally different version of Elixir (listed in dockerfiles). This version is actually a lot older (1.4.5) simply because it hasn't got updated for a while. Anyway, this may only affect performance (if there were any performance upgrades between 1.4.5 -> 1.5.x) since as I said - this is an docker container - Erlang/Elixir in it are only driving MongoosePush and doesn't affect any other environments.

PR #58, opened 2018-03-21
Add 'sound' to APIv2

ludwikbukowski:


lib/mongoose_push.ex
```diff
@@ -37,7 +37,7 @@ defmodule MongoosePush do
 
   Field `:data` may conatin any custom data that have to be delivered to the target device, while
   field `:alert`, if present, must contain at least `:title` and `:body`. The `:alert` field may also
-  contain: `:tag` (option specific to FCM service), `:topic` and `:bagde` (specific to APNS).
+  contain: :sound, `:tag` (option specific to FCM service), `:topic` and `:bagde` (specific to APNS).
```
michalwski, line 40: Do I understand it correctly that the `sound` can be passed in the `alert` parameters doing the HTTP call to MongoosePush? If so it'd worth to document it in README.md file and maybe in the sagger spec which is mentioned in the readme.

ludwikbukowski: @rslota

rslota: Yeah, this should be documented and also tested before merging. I'll add docs and tests as soon I'll be able to.

rslota: And yes, you understand correctly.

PR #59, opened 2018-03-21
Add support for 'sound' and 'priority' options

rslota:
This PR adds support for `sound` and `priority` options for both APSN and FCM.


michalwski (approved):
Looks good to me! I think we can tag new version after merging this or are there other changes possible soon?

rslota:
> Looks good to me! I think we can tag new version after merging this or are there other changes possible soon?

Right now I don't plan to add anything else. I'll make a new release for sure, but I'm also tempted to release it as 1.0.0 (finally). I don't really see the point in waiting, especially considering that APIv2 seems to be stable (we are only adding things).

michalwski:
1.0.0 sounds good to me :) let's do it finally :)

PR #60, opened 2018-03-29
Make baseimage version fixed

rslota:
This PR uses fixed version of docker base image and builds public release using Erlang/Elixir that are configured and available on Travis.

### DIFF
--- a/tools/travis-release.sh
+++ b/tools/travis-release.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+mix local.hex --force
+mix local.rebar --force
+
+export MIX_ENV=prod 
+
+mix do deps.get, release
+tar -czf mongoose_push.tar.gz -C _build/${MIX_ENV}/rel/mongoose_push .
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script: tools/travis-test.sh
 matrix:
     include:
         - stage: test
-          otp_release: 20.0
+          otp_release: 20.3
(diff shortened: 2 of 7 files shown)


michalwski:
I see one of the Travis jobs failed. Is it a random failure?

rslota:
> I see one of the Travis jobs failed. Is it a random failure?

Actually that was a misdiagnosis, so I'll call this PR `wip`. I was sure it's the reason of some failures but apparently it's not.

tools/travis-release.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+mix local.hex --force
+mix local.rebar --force
+
+
+export MIX_ENV=prod
+
+mix do deps.get, deps.compile
```
mongoosepushbot: Use semicolon or linefeed before 'do' (or quote to make it literal).

tools/travis-release.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+mix local.hex --force
+mix local.rebar --force
+
+export MIX_ENV=prod 
+
+mix do deps.get, release
```
mongoosepushbot, line 9: Use semicolon or linefeed before 'do' (or quote to make it literal).

rslota:
> Actually that was a misdiagnosis, so I'll call this PR wip. I was sure it's the reason of some failures but apparently it's not.

@michalwski It's not `wip` anymore. Ready to be merged.

PR #64, opened 2018-04-19
Add support for HTTP/2

rslota:
* HTTP/2 is now default protocol. HTTP 1.1 (over TLS) can still be used without any changes by clients that don't support HTTP/2. It's however strongly recommended to use HTTP/2 capable client while connecting to MongoosePush. 

* This PR also improves a bit timed metrics - adds timers specific to `service`, `mode` and result of the request.

* Min. Elixir version has been changed to `1.5.x` - this is due to some `maru` / `plug` changes - the most recent version of those deps, may not compile on `1.4.x`.

This is also the last PR before `1.0.0` release.


lib/mongoose_push/api/v1.ex
```diff
@@ -32,20 +32,14 @@ defmodule MongoosePush.API.V1 do
...
-            |> json(%{:details => reason})
-          {:error, _reason} ->
-            conn
-            |> put_status(500)
-            |> json(nil)
-        end
+        {status, payload} =
+          device_id
+          |> MongoosePush.push(notification)
+          |> MongoosePush.API.Errors.to_status
```
arkgil: Couldn't you match on the error here? IMO `Errors` module shouldn't handle the success case. Or, you could rename the function, something along the lines of `push_result_to_response`?

mix.exs
```diff
@@ -25,9 +25,13 @@ defmodule MongoosePush.Mixfile do
...
     [
-     {:pigeon, github: "rslota/pigeon"},
-     {:chatterbox, github: "rslota/chatterbox", override: true},
-     {:maru, "~> 0.12", override: true},
+     {:pigeon, github: "rslota/pigeon", ref: "f85b74e"},
+     {:chatterbox, github: "rslota/chatterbox", ref: "4a8abc9", override: true},
+
+     {:maru,  github: "rslota/maru", ref: "54fc038", override: true},
+     {:cowboy,  "~> 2.3", override: true},
+     {:jason, "~> 1.0"}, # This should be included by cowboy, but it's not for some reason
```
arkgil: You can remove the comment here 🙂  It'll be included in the next release of Phoenix, but cowboy has nothing to do with it 🤠

mix.exs
```diff
@@ -4,8 +4,8 @@ defmodule MongoosePush.Mixfile do
   def project do
     [
       app: :mongoose_push,
-      version: "0.10.0",
-      elixir: "~> 1.4",
+      version: "1.0.0",
+      elixir: "~> 1.5",
```
arkgil, line 8: Why not jump straight to 1.6? You're releasing a major version anyway 🙂

lib/mongoose_push/api/v1.ex
arkgil: Ofc such function could be moved to `MongoosePush.API` module

mix.exs
rslota: Because I want to maintain some level of compatibility. I don't want to force anyone to use `1.6` only because I want to. I'm fine with listing here only a minimal Elixir version.

lib/mongoose_push/api/v1.ex
rslota: You're right, I'll move it to `MongoosePush.API`.

mix.exs
rslota: I don't use Phoenix - how is that relevant? And you're right, it's not cowboy's fault, but I'm not sure which lib is at fault here. Cowboy is the one that crashes because of lack of this lib.

arkgil (approved):

PR #65, opened 2018-05-15
Retrun 460 on invalid device token

kanes115:
This PR makes MongoosePush return 401 instead of 500 on invalid device token error.


kanes115:
TODO: Decide on the code and the payload.

kanes115:
Is that what you thought about, @rslota?

rslota:
@kanes115, Thanks for PR!
For now I think this solution is just fine, but we need to create more generic one in future - map all known errors to our custom codes and document them. 
Before we merge this please create an issue to do just that. Also, please change the PR name (error code has changed).

kanes115:
@rslota, we could think of a more generic way when collected more errors we want to inform client about. For now it is just one custom code so I thought it's pointless to abstract it. Thanks!

PR #70, opened 2018-06-20
Add MongoosePushRuntimeTest

rslota:
This new test suites checks whether runtime system supports all necessary features. Right now this only tests TLS cipher suites required by APNS as this is common issue.

### DIFF
--- a/test/mongoose_push_runtime_test.exs
+++ b/test/mongoose_push_runtime_test.exs
@@ -1,0 +1,24 @@
+defmodule MongoosePushRuntimeTest do
+  use ExUnit.Case
+
+  test "tls ciphers required by apns are available" do
+    # APNS supports only:
+    # - ECDHE-RSA-AES128-GCM-SHA256
+    # - ECDHE-RSA-AES256-GCM-SHA384
+
+    all_ciphers = :ssl.cipher_suites()
+    apns_ciphers = Enum.filter(all_ciphers, fn cipher ->
+      case cipher do
+        {:ecdhe_rsa, :aes_128_gcm, _, :sha256} ->
+          true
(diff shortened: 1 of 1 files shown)


test/mongoose_push_runtime_test.exs
```diff
@@ -0,0 +1,24 @@
...
+        {:ecdhe_rsa, :aes_128_gcm, _, :sha256} ->
+          true
+        {:ecdhe_rsa, :aes_256_gcm, _, :sha384} ->
+          true
+        _ ->
+          false
+      end
+    end)
+
+    assert length(apns_ciphers) > 0
```
michalwski, line 21: Is there a built-in assertion checking if a element is on a list? The test case would be simpler, shorter and easier to reason about if we call such assertion twice with different cipher suite name. Sth like the following:
```elixir
assert_in_list :aes_128_gcm all_ciphers
assert_in_list :aes_256_gcm all_ciphers
```

michalwski (commented):
I had one question to the test.

test/mongoose_push_runtime_test.exs
rslota: Three issues:
 - No, there is no such macro, at least in standard lib
 - We need to check whether at least one of those ciphers is available, not both as in your example
 - We need to match 3 of 4 elements in tuple, so even if we ignore above points, at least map with double-head would be needed to strip 4th element of tuple and also handle tuples that already have only 3 elements.

michalwski: Fair enough, thanks for explanation.

rslota: I really wanted to make it more readable, but I don't see better option that is functionally the same.

PR #84, opened 2019-04-05
Allow sending time_to_live field

kanes115:
This PR relies on merging [this PR](https://github.com/rslota/pigeon/pull/1) to Pigeon.

This PR lets mongooseim send `time_to_live` option in notifications.

### DIFF
--- a/test/mongoose_push_test.exs
+++ b/test/mongoose_push_test.exs
@@ -184,7 +184,7 @@ defmodule MongoosePushTest do
       assert notification.alert[:tag] == fcm_data["tag"]
       assert notification.alert[:sound] == fcm_data["sound"]
       assert notification[:data] == fcm_custom
-      assert notification[:time_to_live] == fcm_data["request_data"]["time_to_live"]
+      assert notification[:time_to_live] == fcm_request["request_data"]["time_to_live"]
     end
   end
 
--- a/test/mongoose_push_test.exs
+++ b/test/mongoose_push_test.exs
@@ -147,7 +147,7 @@ defmodule MongoosePushTest do
         tag: string(min: 3, max: 15, chars: :ascii),
         sound: string(min: 3, max: 15, chars: :ascii),
         click_action: string(min: 3, max: 15, chars: :ascii),
-        time_to_live: int(min: 0, max: 4500),
+        time_to_live: int(min: 0, max: 2419200),
         priority: choose(from: [value(:normal), value(:high)]),
       ], repeat_for: 10 do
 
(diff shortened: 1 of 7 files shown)


test/mongoose_push_test.exs
```diff
@@ -147,12 +147,14 @@ defmodule MongoosePushTest do
...
         sound: string(min: 3, max: 15, chars: :ascii),
         click_action: string(min: 3, max: 15, chars: :ascii),
+        time_to_live: int(min: 0, max: 4500),
         priority: choose(from: [value(:normal), value(:high)]),
       ], repeat_for: 10 do
 
       notification =
         %{:service => :fcm,
           :priority => priority,
+          :time_to_live => time_to_live,
```
rslota, line 157: There is no actual check whether this is correctly set in the notification. We need that tested.

lib/mongoose_push.ex
```diff
@@ -63,6 +63,7 @@ defmodule MongoosePush do
         request
         |> Map.put(:alert, request[:alert])
         |> Map.put(:data, request[:data])
+        |> Map.put(:time_to_live, request[:time_to_live])
```
rslota: Why do we need it here? `alert` and `data` need this to ensure those keys are in `request` map, but I don't see any logic in this PR that would require `time_to_live` to be present regardless of whether it's set or not.

rslota (changes requested):
The code looks good in general. Tests and documentation (in `README.md`) seem to be missing though.

test/mongoose_push_test.exs
```diff
@@ -182,6 +184,7 @@ defmodule MongoosePushTest do
       assert notification.alert[:tag] == fcm_data["tag"]
       assert notification.alert[:sound] == fcm_data["sound"]
       assert notification[:data] == fcm_custom
+      assert notification[:time_to_live] == fcm_data["time_to_live"]
```
rslota: According to FCM documentation, `time_to_live` should be at the top level, among things like `priority`. This means that this check should be: 

```
assert notification[:time_to_live] == fcm_request["request_data"]["time_to_live"]
```

rslota (changes requested):
The code looks good to me, but the documentation in `README.md` is still missing. We should add this "field" to list of supported options with a note regarding APNS incompatibility.

kanes115:
@rslota Hope this kind of note is sufficient (the best source is the FCM documentation in this case, isn't it?)

rslota (approved):

PR #85, opened 2019-04-05
Add support for OTP 21.3

rslota:
* Dependencies have been updated so that project compiles with Erlang/OTP 21.3
* Travis matrix has been updated to test and deploy with Erlang/OTP 21.3
* Travis matrix has been updated to test and deploy with Elixir 1.7.x
  (Elixir 1.8.x is temporarly incompatible due to Travis issues with the
  build)
* Version of MongoosePush changes to 1.0.2 as the official Docker image
  is now based on newer Elixir/Erlang platform


kanes115 (approved):
I think this would be worth noting somewhere (in README.md for example) that there are problems with versions of erlang 20.3.X (where X < 2) to save future contributors' time. It is not so obvious to deduce that there are CIPHER suites problems in TLS handshake with APNS.

rslota:
> I think this would be worth noting somewhere (in README.md for example) that there are problems with versions of erlang 20.3.X (where X < 2) to save future contributors' time. It is not so obvious to deduce that there are CIPHER suites problems in TLS handshake with APNS.

Thanks for the suggestion! I've added a small note to `README.md` that explains the issue.

PR #88, opened 2019-06-18
Add date to logger output

### LINKED ISSUE #87: Date is missing in logs
Each logs adds only time of day, date is missing which is not the expected result.

rslota:
Resolves #87 .


config/config.exs
```diff
@@ -29,6 +29,29 @@ config :plug, :statuses, %{
...
+lager_formater_config = [:date, 'T', :time, :color, ' [', :severity, '] ', :pid, ' ', :message, '\e[0m\r\n']
+config :lager,
+  colored: true,
+  handlers: [
+    lager_console_backend: [
+      level: :info,
+      formatter: :lager_default_formatter,
+      formatter_config: lager_formater_config
+    ],
+    lager_file_backend: [
```
arkgil, line 41: Why the split between two log files? Does it mimic the default lager behaviour?

rslota: Yes, the goal here was to copy what `lager` is doing by default.

arkgil (approved):

PR #89, opened 2019-06-19
Hide FCM key from logs

rslota:
* FCM server key is now partially hidden in logs for added security
* Some `Enum.into` have been changed to `Keyword.merge` as per compiler
  warning suggestion.


arkgil (approved):

arkgil:
Looks good! Although we need to make the CI tests pass before merge 😅

rslota:
> Looks good! Although we need to make the CI tests pass before merge 😅

The CI that is complaining is CircleCI that is not configured yet for this project. I failed to disable error reporting before I posted PRs.

PR #92, opened 2019-06-27
Disable colored logs

arkgil:
ANSI color codes make it harder to parse logs and extract structured information from them.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -31,7 +31,7 @@ config :plug, :statuses, %{
 
 lager_formater_config = [:date, 'T', :time, :color, ' [', :severity, '] ', :pid, ' ', :message, '\e[0m\r\n']
 config :lager,
-  colored: true,
+  colored: false,
   handlers: [
     lager_console_backend: [
       level: :info,


config/config.exs
```diff
@@ -29,7 +29,7 @@ config :plug, :statuses, %{
   460 => "Invalid device token"
 }
 
-lager_formater_config = [:date, 'T', :time, :color, ' [', :severity, '] ', :pid, ' ', :message, '\e[0m\r\n']
+lager_formater_config = [:date, 'T', :time, ' [', :severity, '] ', :pid, ' ', :message, '\e[0m\r\n']
 config :lager,
   colored: true,
```
rslota:
```suggestion
  colored: false,
```

rslota: Doesn't the change above fix the issue instead? If this doesn't remove the escape sequence, I still think we should do both.

rslota (changes requested):

config/config.exs
arkgil: @rslota done :)

rslota (approved):

PR #96, opened 2019-07-18
Change APNS to work with Sparrow instead of Pidgeon

kmakiela:
* Changed supervision tree. Sparrow exports one pool supervisor while pidgeon operates on lists of workers, so there are now new supervisors added for both services
* Added necessary files and modules. In the process some modules were reorganised and specific functions changed or deleted
* Reorganised modules that were no longer needed for APNS to be FCM-specific
* Changed tests to work with new version. There were a few imcompatibilities in tests and some of them were no longer valid (i.e. testing active workers)
* Fixed a few inconsistencies with specs
* Fixed a few typos


lib/mongoose_push.ex
```diff
@@ -55,18 +63,18 @@ defmodule MongoosePush do
   @spec push(String.t(), request) :: :ok | {:error, term}
   def push(device_id, %{:service => service} = request) do
     mode = Map.get(request, :mode, :prod)
-    worker = Pools.select_worker(service, mode)
     module = MongoosePush.Application.services()[service]
+    [pool | _] = Pools.pools_by_mode(service, mode)
```
rslota: This doesn't look good here - We're using FCM specific function to get pool name for generic service. 
This call (pool selection) should probably be a service-provided function, where, FCM uses this one, while APNS that uses Sparrow, should use https://github.com/esl/sparrow/blob/b41ea2cc9e6c97ff5db784cb49402ec87234295d/lib/sparrow/pools_warden.ex#L40.

lib/mongoose_push/application.ex
```diff
@@ -48,17 +48,17 @@ defmodule MongoosePush.Application do
     ]
   end
 
-  defp workers do
+  defp create_pool_supervisor do
```
rslota: I would suggest to call it `children` to keep it consistent with Supervisor's naming and also to keep it generic.

lib/mongoose_push/application.ex
```diff
@@ -13,7 +13,7 @@ defmodule MongoosePush.Application do
     set_loglevel(loglevel)
 
     # Define workers and child supervisors to be supervised
-    children = List.flatten(workers())
+    children = List.flatten([create_pool_supervisor()])
```
rslota: Why do we wrap the call in on-element list? It seems to be redundant. 
Also, since we expect each service to return exactly one entry, the `List.flatten` is also redundant.

lib/mongoose_push/application.ex
```diff
@@ -48,17 +48,17 @@ defmodule MongoosePush.Application do
...
-      Enum.map(pools_config, &module.workers/1)
+      module.create_supervisor(pools_config)
     end
   end
 
   defp ensure_mode(config) do
     case config[:mode] do
       nil ->
-        Enum.into([mode: mode(config)], config)
+        Keyword.merge([mode: mode(config)], config)
```
rslota: We need to switch the arguments, as in `Keyword.merge` it's the second argument that overrides first one in case of conflict. I know that given context, the conflict is not possible, but in case of any copy-paste, it makes more sense to have "added" part as second argument.

lib/mongoose_push/service.ex
```diff
@@ -7,7 +7,7 @@ defmodule MongoosePush.Service do
 
   @callback push(Service.notification(), String.t(), atom(), Service.options()) ::
               :ok | {:error, term}
-  @callback prepare_notification(String.t(), MongoosePush.request()) ::
+  @callback prepare_notification(String.t(), MongoosePush.request(), String.t()) ::
               Service.notification()
-  @callback workers({atom, Keyword.t()} | nil) :: list(Supervisor.Spec.spec())
+  @callback create_supervisor({atom, Keyword.t()} | nil) :: {module(), term()}
```
rslota: I would suggest calling this function `supervisor_entry` to indicate, that it returns an `entry` for Supervisor. `create_supervisor` suggest 2 things: 
* that it "creates" something, which is not true, it only returns a spec
* that the thing being created is a supervisor, which may be true, but is not relevant - this could be a worker and it could be just fine

lib/mongoose_push/service.ex
```diff
@@ -7,7 +7,7 @@ defmodule MongoosePush.Service do
 
   @callback push(Service.notification(), String.t(), atom(), Service.options()) ::
               :ok | {:error, term}
-  @callback prepare_notification(String.t(), MongoosePush.request()) ::
+  @callback prepare_notification(String.t(), MongoosePush.request(), String.t()) ::
```
rslota: I'm pretty sure, that pool name is actually an atom. Can we verify that?

lib/mongoose_push/service/apns.ex
```diff
@@ -103,36 +84,19 @@ defmodule MongoosePush.Service.APNS do
...
+    do: Notification.add_mutable_content(notification)
 
   defp maybe_mutable_content(notification, _), do: notification
+
+  defp maybe(notification, _function, nil), do: notification
+  defp maybe(notification, function, arg), do: apply(Notification, function, [notification, arg])
+
+  defp maybe_add_data(notification, nil), do: notification
+
+  defp maybe_add_data(notification, arg) do
```
rslota: Does it need to be a "maybe" type function? If the `:data` is nil, we could just fold over an empty map and would produce the same result.

lib/mongoose_push/service/apns/state.ex
```diff
@@ -0,0 +1,57 @@
...
+  require Logger
+  alias MongoosePush.Service.APNS.Certificate
+
+  def start_link(arg) do
+    GenServer.start_link(__MODULE__, arg)
+  end
+
+  @impl true
+  def init([pool_configs]) do
+    default_topics =
```
rslota: I would suggest moving this logic to external function that could be called here. This function could be then also be used to reload the data in question. This function should parse configs and insert topics into the `ets` table.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  @spec start_link([Keyword.t()]) :: Supervisor.on_start()
+  def start_link(arg) do
+    Supervisor.start_link(__MODULE__, arg)
+  end
+
+  @impl true
+  def init(pool_configs) do
+    {dev_list, prod_list} = List.foldl(pool_configs, {[], []}, &convert_pool_to_sparrow/2)
+    sparrow_config = [{:dev, dev_list}, {:prod, prod_list}]
```
rslota: I would suggest moving this whole translation to external function, to keep the init clean and simple.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,70 @@
+defmodule MongoosePush.Service.APNS.Supervisor do
+  @moduledoc """
+  APNS module supervising Sparrow's PoolSupervisor and APNS State
+  """
+  use Supervisor, id: :apns_supervisor
+  require Logger
+
+  @spec start_link([Keyword.t()]) :: Supervisor.on_start()
```
rslota: I feel that we need to create a type that describes in-depth how this configuration looks like (`pools_config`). Previously it was used only within one file only, now it's being passed around through multiple layers, one of which is a generic behavior.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,70 @@
...
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  def convert_pool_to_sparrow({pool_name, pool_config}, {dev_list, prod_list}) do
+    auth_type = {:auth_type, :certificate_based}
+    cert = {:cert, pool_config[:cert]}
+    key = {:key, pool_config[:key]}
+    pool_size = {:worker_num, pool_config[:pool_size]}
+
+    endpoint_mode =
```
rslota: We could move this to a "module tag" as a map. For example:
```
@default_endpoints %{
  dev: "api.development.push.apple.com"
  prod: "api.push.apple.com"
}
```

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,70 @@
...
+    single_config =
+      [auth_type, cert, key, pool_size, endpoint, port, name]
+      |> Enum.filter(fn {_key, value} -> !is_nil(value) end)
+
+    case pool_config[:mode] do
+      :dev ->
+        {dev_list ++ [single_config], prod_list}
+
+      :prod ->
+        {dev_list, prod_list ++ [single_config]}
```
rslota:
```suggestion
        {dev_list, [single_config | prod_list]}
```

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,70 @@
...
+
+    name = {:pool_name, pool_name}
+
+    single_config =
+      [auth_type, cert, key, pool_size, endpoint, port, name]
+      |> Enum.filter(fn {_key, value} -> !is_nil(value) end)
+
+    case pool_config[:mode] do
+      :dev ->
+        {dev_list ++ [single_config], prod_list}
```
rslota:
```suggestion
        {[single_config | dev_list], prod_list}
```

lib/mongoose_push/service/fcm.ex
```diff
@@ -37,7 +38,9 @@ defmodule MongoosePush.Service.FCM do
 
   @spec push(Service.notification(), String.t(), atom(), Service.options()) ::
           :ok | {:error, term}
-  def push(notification, device_id, worker, opts \\ []) do
+  def push(notification, device_id, _pool, opts \\ []) do
+    worker = Pools.select_worker(:fcm, :prod)
```
rslota: The second argument here (`mode`) is redundant, as FCM does not support multiple modes.

lib/mongoose_push/service/fcm/pool_supervisor.ex
```diff
@@ -0,0 +1,49 @@
+defmodule MongoosePush.Service.FCM.Pool.Supervisor do
```
rslota: The name of this file doesn't match the module name. It should be ` lib/mongoose_push/service/fcm/pool/supervisor.ex`.

lib/mongoose_push/service/fcm/pool_supervisor.ex
```diff
@@ -0,0 +1,49 @@
...
+
+            Supervisor.Spec.worker(
+              Pigeon.GCMWorker,
+              [worker_name, pool_config],
+              id: worker_name
+            )
+          end)
+      end
+
+    list = List.flatten(nested_list)
```
rslota: This name, and also the `nested_list` should be more descriptive. "list" doesn't tell anything about the variable other then type.

rslota (changes requested):
Thanks for the PR!
In general code looks nice, but there are some minor structural / style issues. 
I haven't reviewed tests yet, but I'm sending the implementation review as "part one" to speedup process.

lib/mongoose_push.ex
```diff
@@ -55,18 +62,17 @@ defmodule MongoosePush do
   @spec push(String.t(), request) :: :ok | {:error, term}
   def push(device_id, %{:service => service} = request) do
     mode = Map.get(request, :mode, :prod)
-    worker = Pools.select_worker(service, mode)
     module = MongoosePush.Application.services()[service]
-
+    pool = module.choose_pool(service, mode)
```
rslota: Actually, we don't need `service` here - the service we're calling already know what it's suppose to be.

lib/mongoose_push/application.ex
```diff
@@ -6,22 +6,37 @@ defmodule MongoosePush.Application do
...
+  @type apns_key :: :cert | :key | :pool_size | :mode | :endpoint | :use_2197
+
+  @typedoc """
+  In FCM `:key` and `:pool_size` are required and `:mode` has to be either `:dev` or `:prod`
+  """
+  @type fcm_config :: [fcm_key: String.t() | atom | integer]
+  @typedoc """
+  In APNS `:cert`, `:key` and `:pool_size` are required. `:mode` has to be either `:dev` or `:prod`
+  """
+  @type apns_config :: [apns_key: String.t() | atom | integer]
```
rslota: According to docs this is not the thing you probably intended. This means, that we have a keyword list with a key `apns_key` of type `String.t() | atom | integer`.
I think that it needs to be:
`[{apns_key, String.t() | atom | integer}]`

lib/mongoose_push/application.ex
```diff
@@ -6,22 +6,37 @@ defmodule MongoosePush.Application do
...
+
+  @type pool_info :: {atom(), fcm_config | apns_config}
```
rslota: What is the `atom()` for? If this is a pool name, we should have a type 
`pool_name() :: atom()`. Typespecs are part of documentation, and the names matter, sometimes even more then actual definition.

rslota:
```suggestion
  @type pool_definition :: {atom(), fcm_config | apns_config}
```

lib/mongoose_push/service.ex
```diff
@@ -7,7 +7,8 @@ defmodule MongoosePush.Service do
 
   @callback push(Service.notification(), String.t(), atom(), Service.options()) ::
               :ok | {:error, term}
-  @callback prepare_notification(String.t(), MongoosePush.request()) ::
+  @callback prepare_notification(String.t(), MongoosePush.request(), atom()) ::
```
rslota: This also could use a `pool_name()` type I mentioned above.

lib/mongoose_push/service/apns.ex
```diff
@@ -5,77 +5,64 @@ defmodule MongoosePush.Service.APNS do
...
+  end
 
-      Supervisor.Spec.worker(
-        Pigeon.APNSWorker,
-        [worker_config],
-        id: worker_name
-      )
-    end)
+  @spec choose_pool(MongoosePush.service(), MongoosePush.mode()) :: atom() | nil
+  def choose_pool(service, mode) do
```
rslota: I don't think we need `service` argument here, as this is already specific for APNS.

lib/mongoose_push/service/apns/state.ex
```diff
@@ -0,0 +1,61 @@
...
+    {:ok, %{}}
+  end
+
+  defp setup_default_topics(pool_configs) do
+    default_topics =
+      Enum.map(pool_configs, fn {name, config} ->
+        {name, setup_default_apns_topic(config)[:default_topic]}
+      end)
+
+    :ets.new(:apns_state, [:set, :public, :named_table])
```
rslota: If we ever used this function more then once, it would fail. I suggest moving `:ets.new` to `init/1`.

rslota: Let's match the result to make sure, this succeeds.

lib/mongoose_push/service/apns/state.ex
```diff
@@ -0,0 +1,61 @@
...
+    :ets.insert(:apns_state, default_topics)
```
rslota: Let's match the result to make sure, this succeeds.

lib/mongoose_push/service/apns/state.ex
```diff
@@ -0,0 +1,61 @@
...
+  use GenServer
+  require Logger
+  alias MongoosePush.Service.APNS.Certificate
+
+  def start_link(arg) do
+    GenServer.start_link(__MODULE__, arg)
+  end
+
+  @impl true
+  def init([pool_configs]) do
```
rslota: Why is that a one-element list? Is seems redundant.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  defp convert_pool_to_sparrow({pool_name, pool_config}, {dev_list, prod_list}) do
+    auth_type = {:auth_type, :certificate_based}
+    cert = {:cert, pool_config[:cert]}
+    key = {:key, pool_config[:key]}
+    pool_size = {:worker_num, pool_config[:pool_size]}
+
+    endpoint_mode =
+      case pool_config[:mode] do
```
rslota: This case seems redundant, as we can just call: 
`@default_endpoints[pool_config[:mode]]`

lib/mongoose_push/service/fcm/pools.ex
```diff
@@ -18,27 +18,17 @@ defmodule MongoosePush.Pools do
   def worker_name(type, name, num), do: String.to_atom(~s"#{type}_#{name}_#{num}")
 
   @doc "Returns lists of pool names that have selected `:mode` set"
-  @spec pools_by_mode(MongoosePush.service(), MongoosePush.mode()) :: list(atom)
-  def pools_by_mode(:fcm = service, _mode) do
-    config = pools_config(service)
+  @spec pools_by_mode() :: list(atom)
+  def pools_by_mode() do
+    config = pools_config(:fcm)
```
rslota: Let's remove `import MongoosePush.Application` and use `alias MongoosePush.Application` instead for improved readability.

test/mongoose_push_runtime_test.exs
```diff
@@ -1,6 +1,10 @@
 defmodule MongoosePushRuntimeTest do
-  use ExUnit.Case
+  use ExUnit.Case, async: false
 
+  setup do
+    TestHelper.reload_app()
+
```
rslota: Redundant newline.

test/mongoose_push_test.exs
```diff
@@ -1,24 +1,27 @@
...
   @test_token "testdeviceid1234"
   @apns_priority_mapping %{normal: "5", high: "10"}
+  @dev_topic "dev_topic"
+  @prod_topic "prod1_override_topic"
 
   setup do
     reset(:fcm)
     reset(:apns)
+    TestHelper.reload_app()
+
```
rslota: Redundant newline.

rslota (changes requested):

lib/mongoose_push/application.ex
```diff
@@ -6,22 +6,38 @@ defmodule MongoosePush.Application do
   use Application
   require Logger
 
+  @typedoc "Possible keys in both services"
```
rslota: This type doc will be visible for only `fcm_key` in generated documentation which makes the content weird. I guess, we change it for something more FCM specific and add similar one for `apns_key`.

lib/mongoose_push/service.ex
```diff
@@ -2,12 +2,14 @@ defmodule MongoosePush.Service do
...
 
   @type notification :: term
 
   @callback push(Service.notification(), String.t(), atom(), Service.options()) ::
               :ok | {:error, term}
-  @callback prepare_notification(String.t(), MongoosePush.request()) ::
+  @callback prepare_notification(String.t(), MongoosePush.request(), Application.pool_name()) ::
               Service.notification()
-  @callback workers({atom, Keyword.t()} | nil) :: list(Supervisor.Spec.spec())
+  @callback supervisor_entry({atom, Keyword.t()} | nil) :: {module(), term()}
```
rslota: Isn't the first element of this tuple a `Application.pool_name()`?

lib/mongoose_push/service/apns.ex
```diff
@@ -103,36 +90,18 @@ defmodule MongoosePush.Service.APNS do
...
-        Logger.warn(
-          ~s"Unable to extract APNS topic from the #{config[:mode]} certificate " <>
-            "due to: #{inspect(reason)}"
-        )
-
-        config
-    end
-  end
-
   defp maybe_mutable_content(notification, true),
```
rslota: Maybe this `maybe` should be used instead. The only difference is the argument, but we can make a `maybe` version without argument to pass.

lib/mongoose_push/service/apns.ex
```diff
@@ -5,77 +5,64 @@ defmodule MongoosePush.Service.APNS do
...
-      "sound" => alert[:sound]
-    }
-    |> Notification.new(device_id, request[:topic], request[:data])
+    Notification.new(device_id, Map.get(request, :mode, :prod))
+    |> maybe(:add_apns_topic, request[:topic] || default_topic)
+    |> add_data(request[:data])
     |> maybe_mutable_content(request[:mutable_content])
-    |> Notification.put_priority(@priority_mapping[request[:priority]])
+    |> maybe(:add_apns_priority, @priority_mapping[request[:priority]])
+    |> Notification.add_title(alert.title)
```
rslota, line 35: We should group `maybe` calls together, and push all "required" direct calls to the top.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -0,0 +1,72 @@
...
+        id: :apns_pool_supervisor
+      ),
+      {MongoosePush.Service.APNS.State, pool_configs}
+    ]
+
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  defp create_sparrow_config(pool_configs) do
+    {dev_list, prod_list} = List.foldl(pool_configs, {[], []}, &convert_pool_to_sparrow/2)
```
rslota: Now that I think about it, I think it would be much simpler to read to split this into 2 steps:
1. We split the original list into 2: `prod` and `dev.`
2. Then on both lists, separately, we run `convert_pool_to_sparrow`. This way, we don't need this big and not very pleasant case at the end of `convert_pool_to_sparrow`.

rslota (changes requested):
Almost ready to merge - I do have some very minor comments though.

rslota (approved):

PR #98, opened 2019-08-22
Change FCM to work with Sparrow instead of Pigeon

kmakiela:
* Changed FCM Legacy configuration and setup to FCM v1
* Updated documentation
* Added a few custom types to improve readability
* Deleted obsolete tests
* Updated tests for FCM and made them compatible with new mock server
* Removed Pigeon dependency
* Updated travis setup to use new mock server


README.md
```diff
@@ -23,12 +23,14 @@ In case of certificates you need to setup the following directory structure:
       * prod_key.pem - Production APNS app certificate's private key (has to be unencrypted)
       * dev_cert.pem - Development APNS app certificate
       * dev_key.pem - Development APNS app certificate's private key (has to be unencrypted)
+    * fcm/
+      * token.json - Project's FCM token
```
rslota: Google's calls it FCM's "service account JSON file", so maybe let's try to use their naming across the board.

README.md
```diff
@@ -69,7 +71,7 @@ Environmental variables to configure production release:
 
 ##### Settings for FCM service:
 * `PUSH_FCM_ENDPOINT` - Hostname of `FCM` service. Set only for local testing. By default this option points to the Google's official hostname
-* `PUSH_FCM_APP_KEY` - App key token to use with `FCM` service
+* `PUSH_FCM_TOKEN_PATH` - Path to `FCM` token. For details look at **Running from DockerHub** section
```
rslota: How about `PUSH_FCM_APP_FILE`?

README.md
```diff
@@ -142,17 +144,18 @@ Lets take a look at sample `FCM` service configuration:
 ```elixir
 config :mongoose_push, fcm: [
     default: [
-        key: "fake_app_key",
+        tokenfile: "path/to/token.json",
```
rslota: `appfile`? This file does not contain any tokens, only X509 private key. And since this is not *only* key, I think more generic could be used.

rslota (changes requested):
I posted review for only docs right now - the PR is not rebased on `master`, so changes include APNS `sparrow` related changes, thus it's very hard to actually find what has actually changed. Please rebase the branch and ping for another review :)

lib/mongoose_push/service/fcm/pool/supervisor.ex
```diff
@@ -14,36 +15,33 @@ defmodule MongoosePush.Service.FCM.Pool.Supervisor do
...
+      Supervisor.child_spec({Sparrow.FCM.V1.Supervisor, sparrow_config}, id: :fcm_sparrow_supervisor)
+    ]
+    Supervisor.init(children, strategy: :one_for_one)
+
+  end
+
+  defp create_sparrow_config(pool_configs) do
+    # For now only 1 fcm pool may be configured
+    pool_config =
+      List.first(pool_configs)
```
rslota: How about:

```
pool_configs
|> List.first()
|> convert_pool_to_sparrow()
```

tools/travis-setup.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Skip this setup for jobs that don't run exunit
+test "${PRESET}" == "exunit" || exit 0
+
+## Generate fake certs
+mix certs.dev
+
+## Start mocks
+docker run -d -p 2197:2197 mobify/apns-http2-mock-server
+docker run -d -p 443:4000 4001:4001 kmakiela/fcm-mock-server
```
rslota: No EOL (end-of-line) character.

tools/travis-setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
rslota: Why is this file added here? Isn't it already present in repo? I think something went wrong with rebase.

tools/travis-setup.sh
rslota:
```suggestion
docker run -d -p 4001:4001 kmakiela/fcm-mock-server
```

... since we don't use the HTTPS one.

test/test_helper.exs
```diff
@@ -17,11 +17,14 @@ defmodule TimeHelper do
...
 
 defmodule TestHelper do
+  @fcm_http_port 4001
   def reload_app() do
     Application.stop(:mongoose_push)
+    Application.stop(:sparrow)
     Application.unload(:mongoose_push)
     Application.load(:mongoose_push)
     {:ok, _} = Application.ensure_all_started(:mongoose_push)
+    Application.put_env(:goth, :endpoint, "http://localhost:#{@fcm_http_port}")
```
rslota: Let's put that before we start `:mongoose_push` to make sure it's loaded, as `:goth` may in future change so that for example it reads config only on startup.

rslota (changes requested):
The PR looks great and is almost ready to be merged (details in comments). There are some thing that I would like to see changed, but I don't mention it now, since we have to make even bigger changes as part of other tickets before we release MongoosePush 2.0 anyway.
In addition to comments, we need to change project version in mix file to `2.0.0-alpha.1`.

tools/travis-setup.sh
kmakiela: We use it, i.e here:
https://github.com/esl/MongoosePush/blob/da06ae7a76a762d4b21e67b9558dab8e4943e3b2/test/mongoose_push_test.exs#L371
and
https://github.com/esl/MongoosePush/blob/da06ae7a76a762d4b21e67b9558dab8e4943e3b2/.circleci/config.yml#L35
Should we change that?

tools/travis-setup.sh
kmakiela: From what I can see this file was deleted by your commit:
https://github.com/esl/MongoosePush/commit/93646bb4c0fe12cc2aa601bc05bcfe4e74eb70c9#diff-9facedb7a291effd4d811b201c2d7dae
On master branch there is no such file now - that's why it's all green here

tools/travis-setup.sh
rslota: Didn't remember that, but yeah, let's use HTTP only reduce complexity.

lib/mongoose_push/service.ex
```diff
@@ -5,11 +5,8 @@ defmodule MongoosePush.Service do
   alias MongoosePush.Application
 
   @type notification :: term
-  @type options :: [option]
 
-  @typep option :: {:timeout, integer()}
-
-  @callback push(notification(), String.t(), atom(), options()) ::
+  @callback push(Service.notification(), String.t(), Application.pool_name(), Service.options()) ::
```
rslota: Where is `Service.options()` defined?

lib/mongoose_push/service/fcm/pool/supervisor.ex
```diff
@@ -5,7 +5,9 @@ defmodule MongoosePush.Service.FCM.Pool.Supervisor do
   use Supervisor, id: :fcm_pool_supervisor
   require Logger
   alias MongoosePush.Application
-  alias MongoosePush.Service.FCM.Pools
+
+  @default_endpoint "fcm.googleapis.com"
+  @default_port 4000
```
rslota: Isn't that suppose to be 443?

tools/travis-setup.sh
rslota: We have to remove it then and update the file that replaced this one. I don't remember how it's called. We can't just add this file to repo, since it's probably not even used now (as it was deleted).

rslota (approved):

PR #102, opened 2019-09-06
Service enabled/disabled option fix

kmakiela:
* Fixed cased when service was disabled
* Added tests for enabled/disabled services

### DIFF
--- a/lib/mongoose_push/application.ex
+++ b/lib/mongoose_push/application.ex
@@ -41,11 +41,12 @@ defmodule MongoosePush.Application do
   @spec pools_config(MongoosePush.service()) :: [pool_definition]
   def pools_config(service) do
     enabled_opt = String.to_atom(~s"#{service}_enabled")
+    service_config = Confex.get_env(:mongoose_push, service, nil)
 
     pools_config =
-      case Confex.get_env(:mongoose_push, enabled_opt, true) do
+      case Confex.get_env(:mongoose_push, enabled_opt, !is_nil(service_config)) do
         false -> []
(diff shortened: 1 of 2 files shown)


lib/mongoose_push/application.ex
```diff
@@ -43,7 +43,8 @@ defmodule MongoosePush.Application do
     enabled_opt = String.to_atom(~s"#{service}_enabled")
 
     pools_config =
-      case Confex.get_env(:mongoose_push, enabled_opt, true) do
+      case Confex.get_env(:mongoose_push, enabled_opt,
+       !is_nil(Confex.get_env(:mongoose_push, service, nil))) do
```
rslota: Let's move this expression to a variable so that's easier to read.

lib/mongoose_push/application.ex
```diff
@@ -41,9 +41,9 @@ defmodule MongoosePush.Application do
   @spec pools_config(MongoosePush.service()) :: [pool_definition]
   def pools_config(service) do
     enabled_opt = String.to_atom(~s"#{service}_enabled")
-
+    service_config = Confex.get_env(:mongoose_push, service, nil)
     pools_config =
-      case Confex.get_env(:mongoose_push, enabled_opt, true) do
+      case Confex.get_env(:mongoose_push, enabled_opt, !is_nil(service_config)) do
         false -> []
         true -> Confex.fetch_env!(:mongoose_push, service)
```
rslota: Isn't that just `service_config` now?

rslota: Also, build failing ;)

rslota (approved):

PR #103, opened 2019-09-09
APNS token authorization support

kmakiela:
* Enabled APNS token support
* Changed config files to work with both tokens and certificates
* Updated README
* Added test for proper token authorization


README.md
```diff
@@ -23,9 +23,13 @@ In case of certificates and token you need to setup the following directory stru
       * prod_key.pem - Production APNS app certificate's private key (has to be unencrypted)
       * dev_cert.pem - Development APNS app certificate
       * dev_key.pem - Development APNS app certificate's private key (has to be unencrypted)
+      * token.p8 - `APNS` authentication token
     * fcm/
       * token.json - `FCM` service account JSON file
 
+In case you want to use `APNS` token authentication you need to provide token and set `key_id` and `team_id` environmental variables. On how to obtain token and `key_id` read:
```
aleklisi: `key_id` -> `PUSH_APNS_DEV_KEY_ID`
`team_id` -> `PUSH_APNS_DEV_TEAM_ID`
As these are environmental variables you refer to later, right?

README.md
```diff
@@ -78,6 +82,9 @@ Environmental variables to configure production release:
 * `PUSH_APNS_DEV_ENDPOINT` - Hostname of `APNS` service. Set only for local testing. By default this option points to the Apple's official hostname
 * `PUSH_APNS_DEV_CERT` - Path Apple's development certfile used to communicate with `APNS`
 * `PUSH_APNS_DEV_KEY` - Path Apple's development keyfile used to communicate with `APNS`
+* `PUSH_APNS_DEV_KEY_ID` - Key ID generated from Apple's developer console. For details look at **Running from DockerHub** section *required for token authentication*
+* `PUSH_APNS_DEV_TEAM_ID` - TEAM ID generated from Apple's developer console. For details look at **Running from DockerHub** section *required for token authenticaton*
```
aleklisi, line 86: TEAM ID -> Team ID or Key ID -> KEY ID for consistency with previous point

lib/mongoose_push/service/apns/state.ex
```diff
@@ -37,7 +37,7 @@ defmodule MongoosePush.Service.APNS.State do
       # so there may be some incompability issues that we may work on based on failure logs.
       case config[:default_topic] do
         nil ->
-          all_topics = Certificate.extract_topics!(config[:cert])
+          all_topics = Certificate.extract_topics!(config[:auth][:cert])
```
aleklisi, line 40: What is the expected result when auth -> cert does not exist?
If you expect `nil` 
If you expect `KeyError` use `config.auth.cert` instead
Checkout the example:
```elixir
iex(1)> m = %{:a => %{:q => 1}}
%{a: %{q: 1}}
iex(2)> m.a.q
1
iex(3)> m[:a][:q]
1
iex(4)> m.a.b
** (KeyError) key :b not found in: %{q: 1}
iex(4)> m[:a][:b]
nil
```

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -31,23 +31,49 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
-    dev_pools =
+    dev_cert_pools =
       pool_configs
-      |> Enum.filter(fn {_, pool_config} -> pool_config[:mode] == :dev end)
-      |> List.foldl([], &convert_pool_to_sparrow/2)
+      |> Enum.filter(fn {_, pool_config} ->
+        pool_config[:mode] == :dev and
+          pool_config[:auth][:type] == :certificate_based
+      end)
+      |> List.foldl([], &convert_cert_pool_to_sparrow/2)
```
aleklisi: Why fold when you can just map? Simpler code && less reading.
If you really want to use fold than fold and filter in the same function or if you decide to use map which (I recommend) use http://erlang.org/doc/man/lists.html#filtermap-2 instead.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -31,23 +31,49 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
 
-    prod_pools =
+    prod_cert_pools =
       pool_configs
-      |> Enum.filter(fn {_, pool_config} -> pool_config[:mode] == :prod end)
-      |> List.foldl([], &convert_pool_to_sparrow/2)
+      |> Enum.filter(fn {_, pool_config} ->
```
aleklisi, line 50: The same as above.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -31,23 +31,49 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+        pool_config[:mode] == :prod and
```
aleklisi, line 51: If you refactor to map only than extract filtering function as it is used 3 times in row parametrised with pool_config `mode` and `auth type`.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+
+    token = [pool_token, key_id, team_id, p8_file]
+
+    {[single_config | mode_list], [token | tokens]}
+  end
+
+  @chars "ABCDEFGHIJKLMNOPQRSTUVWXYZ" |> String.split("")
+  defp random_atom(len) do
+    1..len
+    |> Enum.reduce([], fn _i, acc ->
```
aleklisi: Check this syntax, do you find it more readable?

```elixir
for _ <- 1..4, do: Enum.random(["A", "B", "C"])
```

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+      key=#{key} or team=#{team}")
+
+      Supervisor.stop(self(), {:error, "Required file does not exist in the given path:
+      p8_file=#{p8_file_path}"})
+    end
+
+    key_id = {:key_id, key}
+    team_id = {:team_id, team}
+    p8_file = {:p8_file_path, p8_file_path}
+    token_id = random_atom(15)
```
aleklisi: If would be better to use String instead of Atom as Strings can be garbage collected and Atoms stay forever in atoms table. http://erlang.org/doc/efficiency_guide/advanced.html#memory

aleklisi: https://youtu.be/_Pwlvy3zz9M?t=1388 I recommend to see the section about `atom table`

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+    pool_token = {:token_id, token_id}
+
+    single_config =
+      [auth_type, pool_token, pool_size, endpoint, port, name]
+      |> Enum.filter(fn {_key, value} -> !is_nil(value) end)
+
+    token = [pool_token, key_id, team_id, p8_file]
```
aleklisi: Maybe provide Struct for the token ??

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+
+    key_id = {:key_id, key}
+    team_id = {:team_id, team}
+    p8_file = {:p8_file_path, p8_file_path}
+    token_id = random_atom(15)
+    pool_token = {:token_id, token_id}
+
+    single_config =
+      [auth_type, pool_token, pool_size, endpoint, port, name]
+      |> Enum.filter(fn {_key, value} -> !is_nil(value) end)
```
aleklisi, line 154: What do you filter it for?

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+
+    token = [pool_token, key_id, team_id, p8_file]
+
+    {[single_config | mode_list], [token | tokens]}
+  end
+
+  @chars "ABCDEFGHIJKLMNOPQRSTUVWXYZ" |> String.split("")
```
aleklisi: How about:
```elixir
for n <- ?A..?Z, do: << n :: utf8 >>
```
??

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+
+    if not File.exists?(p8_file_path) do
+      Logger.error(~s"Required file does not exist in the given path:
+      p8_file=#{p8_file_path}")
+
+      Supervisor.stop(self(), {:error, "Required file does not exist in the given path:
+      p8_file=#{p8_file_path}"})
+    end
+
+    if is_nil(key) or is_nil(team) do
```
aleklisi: Why not doing the full check for all config opts?
Checkout PoC: https://github.com/esl/sparrow/pull/4
I would highly recommend to add check config responsibility to be moved to https://github.com/esl/sparrow and do it on starting sparrow.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+      case pool_config[:use_2197] do
+        true ->
+          2197
+
+        _ ->
+          nil
+      end
+
+    port = {:port, port_config}
+
```
aleklisi: Why do you need the new lines between `port` and `name` and not new lines between `key` and `team`??

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+    if not File.exists?(p8_file_path) do
+      Logger.error(~s"Required file does not exist in the given path:
+      p8_file=#{p8_file_path}")
+
+      Supervisor.stop(self(), {:error, "Required file does not exist in the given path:
+      p8_file=#{p8_file_path}"})
+    end
+
+    if is_nil(key) or is_nil(team) do
+      Logger.error(~s"Required authentication elements are missing:
```
aleklisi: Maybe keep all 3 conditions together within a single if?

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -72,4 +98,69 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+
+  defp convert_token_pool_to_sparrow({pool_name, pool_config}, {mode_list, tokens}) do
+    auth_type = {:auth_type, :token_based}
+    pool_size = {:worker_num, pool_config[:pool_size]}
+
+    endpoint_mode = @default_endpoints[pool_config[:mode]]
+    endpoint = {:endpoint, pool_config[:endpoint] || endpoint_mode}
+
+    port_config =
+      case pool_config[:use_2197] do
```
aleklisi: How about using `if` to make it one liner instead??
```elixir
port_config = if pool_config[:use_2197], do: 2197, else: nil
```

aleklisi (changes requested):
Good PR
1) @goddammit please language check the docs
2) You provide some config check but you do not check when the app fails to start do to wrong config.
3) Use map instead of fold, folds are more general but map is usually simpler.

README.md
```diff
@@ -12,8 +12,8 @@ notifications** to `FCM` (Firebase Cloud Messaging) and/or
 
 #### Running from DockerHub
 
-We provide already built MongoosePush images. If you just want to use it, then all you need is `docker`, `FCM` app token and/or `APNS` app certificates.
-In case of certificates and token you need to setup the following directory structure:
+We provide already built MongoosePush images. If you just want to use it, then all you need is `docker`, `FCM` app token and/or `APNS` app certificates or token. `APNS` can be configured with either of these - you only need one.
+For the full configuration you need to setup the following directory structure:
```
goddammit: For the full configuration you need to set the following directory structure up:

README.md
goddammit: If you want to use the `APNS` token authentication, you need to provide a token and set the `key_id` and `team_id` environmental variables. To see how to obtain token and `key_id` read:

README.md
```diff
@@ -12,8 +12,8 @@ notifications** to `FCM` (Firebase Cloud Messaging) and/or
 
 #### Running from DockerHub
 
-We provide already built MongoosePush images. If you just want to use it, then all you need is `docker`, `FCM` app token and/or `APNS` app certificates.
-In case of certificates and token you need to setup the following directory structure:
+We provide already built MongoosePush images. If you just want to use it, then all you need is `docker`, `FCM` app token and/or `APNS` app certificates or token. `APNS` can be configured with either of these - you only need one.
```
goddammit: We provide prebuilt MongoosePush images.  
Alt:
We provide MongoosePush images you can use. 

Configuration requires either an FCM token, APNS certificates or an APNS token. Depending on your usecase, you can have some or all of them in a standalone MongoosePush instance or using a docker container.

README.md
```diff
@@ -23,9 +23,13 @@ In case of certificates and token you need to setup the following directory stru
       * prod_key.pem - Production APNS app certificate's private key (has to be unencrypted)
       * dev_cert.pem - Development APNS app certificate
       * dev_key.pem - Development APNS app certificate's private key (has to be unencrypted)
+      * token.p8 - `APNS` authentication token
     * fcm/
       * token.json - `FCM` service account JSON file
 
+In case you want to use `APNS` token authentication you need to provide token and set `key_id` and `team_id` environmental variables. On how to obtain token and `key_id` read:
+https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/establishing_a_token_based_connection_to_apns  
+On how to obtain `team_id` read: https://www.mobiloud.com/help/knowledge-base/ios-app-transfer/
```
goddammit: To see how to obtain

lib/mongoose_push/service/apns/state.ex
kmakiela: ~~I expect `KeyError`, it's required~~ That's not true, because we may call this function even with token based auth and then this file may not exist

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: Elixir's `Enum.filter_map` has been deprecated, will change foldl to map though

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: I can't `filter` after `map` because there are different mapping functions for different filter clauses

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: It sure is, I just copied It from Sparrow

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: Are You talking about `token_id`? If that's the case it has to be an atom because Sparrow requires it. Additionally the amount of dynamically created atoms depends on how many pool configurations there are and it's not a lot

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: It would require whole new module just for that - is it worth it? IMO it's pretty clear

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: For options that may be `nil`, e.g. `port` or `endpoint`

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: The PoC is not merged to master in sparrow and that's why @rslota suggested having the checks done in MPush to be sure that valid config is passed to Sparrow

config/dev.exs
```diff
@@ -34,17 +34,23 @@ config :mongoose_push,
...
       endpoint: "localhost",
-      cert: "priv/apns/dev_cert.pem",
-      key: "priv/apns/dev_key.pem",
       mode: :dev,
       use_2197: true,
       pool_size: 5
     ],
     prod: [
+      auth: %{
+        type: :certificate_based,
```
rslota: How about just `:certificate` / `:token` without "_based" suffix? IMHO values in config shouldn't contain text that does not bring any information.

config/prod.exs
```diff
@@ -33,18 +33,30 @@ config :mongoose_push,
 config :mongoose_push,
   apns: [
     dev: [
+      auth: %{
+        type: {:system, :string, "PUSH_APNS_AUTH_TYPE", :token_based},
```
rslota: This should be an `:atom`, unless it's not supported.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -31,45 +31,104 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
 
+    key = pool_config[:auth][:key_id]
+    team = pool_config[:auth][:team_id]
+    p8_file_path = pool_config[:auth][:p8_file_path]
+
+    if is_nil(key) or is_nil(team) or not File.exists?(p8_file_path) do
+      Logger.error(~s"Required authentication elements are missing. Got:
+      key=#{key}, team=#{team}, p8_file=#{p8_file_path}")
+
+      Supervisor.stop(self(), {:error, "Required authentication elements are missing. Got:
```
rslota: Two issues here: 
1. We should return the error and use `{:stop, reason}` return value from `init/1`
2. reason should never be a string - we need something the we can easely match on. The string is logged, which great, and that should be it.

lib/mongoose_push/service/apns/supervisor.ex
rslota: This looks like a constant size list, which is just weird. Only after analyzing the whole function, I noticed, that all elements are 2-tuples, which makes it a keyword list, which is fine. IMHO, this keyword list, should be created here, inline, so that all keys are listed in one place, explicitly and only values are referenced by a variable.

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -31,45 +31,104 @@ defmodule MongoosePush.Service.APNS.Supervisor do
...
+    port_config = if pool_config[:use_2197], do: 2197, else: nil
+    port = {:port, port_config}
+
+    name = {:pool_name, pool_name}
+
+    [auth_type, cert, key, pool_size, endpoint, port, name]
+    |> Enum.filter(fn {_key, value} -> !is_nil(value) end)
+  end
+
+  defp convert_token_pool_to_sparrow({pool_name, pool_config}, {mode_list, tokens}) do
```
rslota: There seem to be a lot of code duplication between those to "convert" functions. Can we extract common parts to another, and keep only auth parts separate?

rslota (changes requested):
I like that we check for options validity when token auth is enabled, but we don't do that for certificates, thus I think, we should add similar check there for consistency. 
Also, I have some other minor comments.
In general though, the PR looks great and covers story scope exactly as it should.

rslota:
@kmakiela Can you please rebase this branch on top of the `master`? I feel that there are unrelated changes here showing.

kmakiela:
@rslota It is rebased on master and git history supports that as well - What unrelated changes are there?

rslota:
@kmakiela there are 4 commits here, one of them is https://github.com/esl/MongoosePush/commit/dbccad584fa7447d0c8fa2ce7daee19ecb3e3b1b which adds pools support.

kmakiela:
@rslota This is the PR You closed today or yesterday - it was a PR for this branch, not master, because master did not have changes from this branch yet and to avoid unnecessary conflicts in the future I decided to create PR here for visibility mostly. Additionally it wouldn't have made sense to create that PR to master because some required changes were included in this PR. Hope that clarifies it

rslota:
@kmakiela ,

Please bump the version of MP to `2.0.0-alpha2` and that we're using the latest version of `sparrow`. Otherwise this PR is ready to be merged in!

rslota (approved):
LGTM!

PR #111, opened 2019-10-31
Add docker compose for setting up mocks

ludwikbukowski:


docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3'
```
rslota: I would suggest moving this file to `test` directory, as it's used mostly for tests and this could help indicate that. 
It would also be great to add some info to ready on this topic (how to easly run tests using this).

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+
+services:
+        fcm-mock:
+                image: kmakiela/fcm-mock-server
```
rslota: Does this indentation have to be that big? :)

docker-compose.yml
ludwikbukowski: I was thinking about `test/docker` dir

rslota: also works :)

ludwikbukowski:
@rslota done

README.md
```diff
@@ -115,6 +115,23 @@ _build/prod/rel/mongoose_push/bin/mongoose_push console
...
+### Running tests
+
+Setup FCM and APNS mocks first:
+```bash
+$ docker-compose -f test/docker/docker-compose.yml up -d
+```
+
+Run tests:
+```bash
+$ mix test
```
rslota, line 132:
```suggestion
$ mix do certs.dev, test
```

rslota: Or maybe even split that with a remark that `certs.dev` is only needed once.

ludwikbukowski: yeah, right - I forgot about certs

rslota (approved):
Looks great, thanks!

PR #113, opened 2019-11-05
Test error codes

zofpolkowska:


test/api/v2_fcm_test.exs
```diff
@@ -0,0 +1,106 @@
...
+        :tag => "tag value"}}
+
+    mock_fcm([%{device_token: "f534534543", status: 404, reason: reason}])
+
+    assert {500, reason} = post(@url, args)
+  end
+
+  test "push to fcm with id mismatch fails" do
+    reason = "SENDER_ID_MISMATCH"
+    args = %{
```
rslota: Those all seem to be the same - maybe we can extract this payload to a function?

test/api/v2_apns_test.exs
```diff
@@ -0,0 +1,152 @@
+defmodule MongoosePushAPIV2APNSTest do
+  require Logger
+  use ExUnit.Case, async: false
+  use Quixir
```
rslota: We don't seem to use it here.

test/api/v2_apns_test.exs
```diff
@@ -0,0 +1,152 @@
...
+  import MongoosePushSupportAPI
```
rslota: If we really have to import something then let's import each function explicitly. Though IMHO it's still better to just use aliases, like e.g.:
```
alias MongoosePushSupportAPI, as: Tools
```

test/support/api.ex
```diff
@@ -0,0 +1,80 @@
+defmodule MongoosePushSupportAPI do
```
rslota:
```suggestion
defmodule MongoosePush.Support.API do
```
(dots in between name segments)

test/api/v2_fcm_test.exs
```diff
@@ -0,0 +1,106 @@
+defmodule MongoosePushAPIV2FCMTest do
```
rslota:
```suggestion
defmodule MongoosePush.API.V2FCMTest do
```

test/api/v2_apns_test.exs
```diff
@@ -0,0 +1,152 @@
+defmodule MongoosePushAPIV2APNSTest do
```
rslota:
```suggestion
defmodule MongoosePush.API.V2APNSTest do
```

rslota (commented):
Looks great in general! I left only some minor comments. Also, please reformat the code with `mix format` - I'm not sure why CI haven't validated that.

test/support/api.ex
```diff
@@ -0,0 +1,91 @@
...
+
+  def get_connection(:apns) do
+    :h2_client.start_link(:https, 'localhost', 2197, [])
+  end
+
+  def get_connection(:fcm) do
+    :h2_client.start_link(:https, 'localhost', 4000, [])
+  end
+
+  def headers(method, path, payload \\ "") do
```
rslota: The default value here already gave us a day of debugging recently (payload was not being passed to this function, but was being sent, thus size of payload was incorrect) - let's remove the default value here, and always pass the payload that we're actually sending. Even if payload is empty, let's make a variable that we pass both here and to `send_request`.

test/support/api.ex
```diff
@@ -0,0 +1,90 @@
...
+      {":path", path},
+      {"content-length", "#{byte_size(payload)}"},
+      {"content-type", "application/json"}
+    ]
+  end
+
+  def get_response(conn) do
+    receive do
+      {:END_STREAM, stream_id} ->
+        :h2_client.get_response(conn, stream_id)
```
rslota: I know that this code is taken from old test suite, but since we're on it, let's fix most glaring issues - we don't check error code here - we may see random failures in test and not be even aware that mock we're talking to refuses to work. We really should return from this function error code and response body, and match on error code in the functions that actually use it.

rslota: I saw there was a problem with dialyzer here - if possible I would give it few minutes, and if it doesn't seem like something that is possible to fix, I would just add this file to dialyzer ignore list.

zofpolkowska: Should we match on the possible values returned by `get_response/2` and log when an error occurs?

rslota: I would suggest to just match and crash. This is an edge case during test setup - we should just know about it before any other side effects occur.

rslota (approved):
LGTM!

PR #114, opened 2019-11-06
Update `dev` release docs / configs

rslota:
* Make `config/dev.exs` connect by default to local mocks that can be
setup the same way as for tests.
* README.md now includes information aboout building `dev` release that
connects to local APNS / FCM mocks.


config/dev.exs
```diff
@@ -42,7 +44,8 @@ config :mongoose_push,
       endpoint: "localhost",
       mode: :dev,
       use_2197: true,
-      pool_size: 5
+      pool_size: 5,
+      tls_opts: []
```
rslota, line 48: This option is here, as recently MongoosePush started validating APNS / FCM server certs by default. As for dev release we expect to be connecting to mocks, we need to disable the verification, as otherwise the release won't boot.

zofpolkowska: Maybe we can add this option to the list here: https://github.com/esl/MongoosePush/commit/7983a853e7993ceca8205687c3397bb596aff1ea#diff-04c6e90faac2675aa89e2176d2eec7d8R194 ?

zofpolkowska (approved):
Now it is much clearer, thank you for the update:)

PR #115, opened 2019-11-06
API V3

kmakiela:
* Reordered files for different versions of API
* Added translation for API V3

### DIFF
--- a/lib/mongoose_push/api/v1/h2_handler.ex
+++ b/lib/mongoose_push/api/v1/h2_handler.ex
@@ -1,0 +1,12 @@
+defmodule MongoosePush.API.V1.H2Handler do
+  @moduledoc """
+    Module for handling internal responses to V1 HTTP2 codes
+  """
+  @behaviour MongoosePush.API
+
+  @spec to_status(:ok | {:error, term}) ::
+    {non_neg_integer, %{details: atom | String.t()} | nil}
+  def to_status(return_val) do
+    MongoosePush.API.V2.H2Handler.to_status(return_val)
+  end
+end
--- a/lib/mongoose_push/api/v1/response_encoder.ex
+++ b/lib/mongoose_push/api/v1/response_encoder.ex
@@ -1,4 +1,4 @@
-defmodule MongoosePush.API.V1.H2Handler do
+defmodule MongoosePush.API.V1.ResponseEncoder do
   @moduledoc """
     Module for handling internal responses to V1 HTTP2 codes
   """
@@ -7,6 +7,6 @@ defmodule MongoosePush.API.V1.H2Handler do
   @spec to_status(:ok | {:error, term}) ::
     {non_neg_integer, %{details: atom | String.t()} | nil}
   def to_status(return_val) do
-    MongoosePush.API.V2.H2Handler.to_status(return_val)
+    MongoosePush.API.V2.ResponseEncoder.to_status(return_val)
   end
 end
--- a/lib/mongoose_push/api/v1/response_encoder.ex
+++ b/lib/mongoose_push/api/v1/response_encoder.ex
@@ -1,11 +1,13 @@
 defmodule MongoosePush.API.V1.ResponseEncoder do
   @moduledoc """
-    Module for handling internal responses to V1 HTTP2 codes
+  Module for handling internal responses to V1 HTTP2 codes
   """
   @behaviour MongoosePush.API
+  alias MongoosePush.Service
 
-  @spec to_status(:ok | {:error, term}) ::
-    {non_neg_integer, %{details: atom | String.t()} | nil}
+  @spec to_status(:ok | {:error, Service.error()} | {:error, MongoosePush.error()}) ::
+          {non_neg_integer, %{details: atom | String.t()} | nil}
+
   def to_status(return_val) do
     MongoosePush.API.V2.ResponseEncoder.to_status(return_val)
   end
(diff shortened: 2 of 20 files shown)


lib/mongoose_push/api/v1/h2_handler.ex
```diff
@@ -0,0 +1,12 @@
+defmodule MongoosePush.API.V1.H2Handler do
```
rslota: This module is not specific to HTTP/2 - HTTP 1.1 is also supported here. I would prefer to name this module something like e.g. `ResponseEncoder`.

lib/mongoose_push/api/v3/h2_handler.ex
```diff
@@ -0,0 +1,52 @@
...
+      case type do
+        :invalid_request ->
+          461
+        :auth ->
+          462
+        :unregistered ->
+          463
+        :unspecified ->
+          464
+        :service_internal ->
```
rslota: `service_internal`, `auth`, both should be 5xx, as they are not in any way connected to the user's input. The auth one is all about auth issues between MongoosePush and Push Service, not user-input auth.

zofpolkowska: `500` for `service_internal` and `511` for `auth` then ?

rslota: Both would be very misleading to client. 
`500` means that it's the server client is connecting to misbehaving which is not true `service_internal` means that 3rd party service APNS / FCM has issues. 
`511` is some weird version of `401`used for network / proxy services. Either way, it suggest that client is missing authorization which is not true at all - `auth` is an error category that all mean that MongoosePush itself is unable to authorize with APNS / FCM. This is normally a configuration error. 

There are no official HTTP codes that cover those cases. In general, what we need to do is to either pick the closest that makes sense (`503` here for both cases) and return well defined details in payload, or go with custom ones. 

Initially for MongoosePush, I figured that the amount of possible errors may be better of with being displayed as custom HTTP codes, but now that I think about it, we can scrap this idea. Thus, feel free to change all those codes to standard codes that correspond closely to our error reasons, group them and let's return well defined `%{reason => "internal_error_code"}` JSON payload.

lib/mongoose_push/api/v3/response_encoder.ex
```diff
@@ -0,0 +1,44 @@
...
+        :payload_too_large -> 413
+        :too_many_requests -> 429
+        :auth -> 503
+        :service_internal -> 503
+        :internal_config -> 503
+        :unspecified -> 520
+        :generic -> 400
+      end
+
+    {return_code, %{:reason => type}}
```
rslota: If we're renaming this field to `reason`, it should be done in the whole V3. 
Also, specs need to be updates (along with behaviour that needs to be relaxed).

lib/mongoose_push/api/v3/response_encoder.ex
```diff
@@ -0,0 +1,44 @@
...
+      case type do
+        :invalid_request -> 400
+        :unregistered -> 410
+        :payload_too_large -> 413
+        :too_many_requests -> 429
+        :auth -> 503
+        :service_internal -> 503
+        :internal_config -> 503
+        :unspecified -> 520
+        :generic -> 400
```
rslota: That's not really true though. Only one specific value here should be 400 - `{:generic, :no_matching_pool}` (we can btw. send `reason => :no_matching_pool` in such case - `:generic` errors are controlled by MongoosePush, so we can match and actually use the `reason`).
All other values should be 500 for now. When we add proper handling for connection issues, we'll get also 503 from this.

rslota (commented):
We're missing tests here - when we introduce API version, we should have a function tests that verify this API. I know that V3 is the same as V2, but the code is copied, not referenced. And event if it was, from tests perspective this is a new thing, that we need to ensure is working properly. For V2 we have `RestV2Test` and we need similar one for V3.

lib/mongoose_push/api/v3/response_encoder.ex
```diff
@@ -0,0 +1,45 @@
...
+        {:internal_config, _} -> {503, type}
+        {:unspecified, _} -> {520, type}
+        {:generic, :no_matching_pool} -> {400, reason}
+        {:generic, _} -> {500, reason}
+      end
+
+    {status, %{:reason => error_reason}}
+  end
+
+  def to_status({:error, reason}) when is_atom(reason) do
```
rslota: Let's remove this case. According to spec and out current implementation, this cannot happen. Thus, let's just crash here, as we'll probably crash earlier anyway.

lib/mongoose_push/api/v3/response_encoder.ex
```diff
@@ -0,0 +1,45 @@
+defmodule MongoosePush.API.V3.ResponseEncoder do
+  @moduledoc """
+  Module for handling internal responses to V3 HTTP2 codes
+  """
+  @behaviour MongoosePush.API
+  alias MongoosePush.Service
+
+  @spec to_status(:ok | {:error, Service.error()} | {:error, MongoosePush.error()}) ::
+          {non_neg_integer, %{details: atom | String.t()} | nil}
```
rslota:
```suggestion
          {non_neg_integer, %{reason: atom | String.t()} | nil}
```

lib/mongoose_push/api/v1/response_encoder.ex
```diff
@@ -0,0 +1,14 @@
+defmodule MongoosePush.API.V1.ResponseEncoder do
+  @moduledoc """
+  Module for handling internal responses to V1 HTTP2 codes
+  """
+  @behaviour MongoosePush.API
+  alias MongoosePush.Service
+
+  @spec to_status(:ok | {:error, Service.error()} | {:error, MongoosePush.error()}) ::
+          {non_neg_integer, %{details: atom | String.t()} | nil}
```
rslota, line 9: This spec is invalid - depending on API version, the key may be different. We should probably relax that to just a map.

rslota (approved):

PR #116, opened 2019-11-08
API v3 doc

zofpolkowska:
Description of the error codes in:
- swagger.json
- readme

### DIFF
--- a/lib/mongoose_push/api/v3.ex
+++ b/lib/mongoose_push/api/v3.ex
@@ -10,6 +10,20 @@ defmodule MongoosePush.API.V3 do
     parsers: [:urlencoded, :json, :multipart]
   )
 
+  desc "description" do
+    detail "details of described request"
+
+    responses do
+      status 200, desc: "OK"
+      status 400, desc: "{"reason" : "invalid_request"|"no_matching_pool"} - the request was invalid"
+      status 410, desc: "{"reason" : "unregistered"} - the device was not registered"
+      status 413, desc: "{"reason" : "payload_too_large"} - the payload was too large"
+      status 429, desc: "{"reason" : "too_many_requests"} - there were too many requests to the server"
+      status 503, desc: "{"reason" : "service_internal"|"internal_config"|"unspecified"} - the internal service or configuration error occured"
+      status 520, desc: "{"reason" : "unspecified"} - the unknown error occured"
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


lib/mongoose_push/api/v3.ex
```diff
@@ -0,0 +1,70 @@
...
+    detail "details of described request"
+
+    responses do
+      status 200, desc: "OK"
+      status 400, desc: "The request was invalid"
+      status 410, desc: "The device was not registered"
+      status 413, desc: "The payload was too large"
+      status 429, desc: "There were too many requests to the server"
+      status 503, desc: "The internal error occured"
+      status 520, desc: "The unknown error occured"
```
rslota: What about 500?

lib/mongoose_push/api/v3.ex
```diff
@@ -0,0 +1,70 @@
...
+  desc "description" do
+    detail "details of described request"
+
+    responses do
+      status 200, desc: "OK"
+      status 400, desc: "The request was invalid"
+      status 410, desc: "The device was not registered"
+      status 413, desc: "The payload was too large"
+      status 429, desc: "There were too many requests to the server"
+      status 503, desc: "The internal error occured"
```
rslota: There are several different error codes that are returned in response body - we should describe them here. 503 is not the only one with this "issue" I think.

lib/mongoose_push/api/v3.ex
```diff
@@ -0,0 +1,70 @@
...
+    detail "details of described request"
+
+    responses do
+      status 200, desc: "OK"
+      status 400, desc: "{"reason" : "invalid_request"|"no_matching_pool"} - the request was invalid"
+      status 410, desc: "{"reason" : "unregistered"} - the device was not registered"
+      status 413, desc: "{"reason" : "payload_too_large"} - the payload was too large"
+      status 429, desc: "{"reason" : "too_many_requests"} - there were too many requests to the server"
+      status 503, desc: "{"reason" : "service_internal"|"internal_config"|"unspecified"} - the internal service or configuration error occured"
+      status 520, desc: "{"reason" : "unspecified"} - the unknown error occured"
```
rslota, line 23: We're still lacking docs for `500` error - we do use that explicitly.

rslota (approved):

PR #124, opened 2019-11-26
Zp/conn errors

zofpolkowska:


rslota (changes requested):
The PR looka good, but it should also include:

- Version bump to `2.0.0-beta.1` (feature complete, only bugfixes left)
- Sparrow update to version that properly returns errors tested here

In other words this is blocked by https://github.com/esl/sparrow/pull/30 .

rslota (approved):

PR #126, opened 2019-11-27
Integration tests

zofpolkowska:


config/integration.exs
```diff
@@ -0,0 +1,123 @@
+use Mix.Config
+
+config :mongoose_push, loglevel: :debug
+
+config :elixometer,
```
rslota: Do we need elixometer for those tests?

config/integration.exs
```diff
@@ -0,0 +1,123 @@
...
+  reporter: :exometer_report_tty,
+  env: Mix.env(),
+  metric_prefix: "mongoose_push"
+
+config :maru, MongoosePush.Router,
```
rslota: I don't think we need to configure `maru` as we don't start any applications for those tests (or at least we shouldn't). Actually we don't need anything in this config right now :)

config/prod.exs
```diff
@@ -21,10 +23,14 @@ config :mongoose_push, fcm_enabled: {:system, :boolean, "PUSH_FCM_ENABLED", true
 
 config :mongoose_push, apns_enabled: {:system, :boolean, "PUSH_APNS_ENABLED", true}
 
+config :mongoose_push,
```
rslota, line 26: We should document existence of this option.

config/prod.exs
```diff
@@ -21,10 +23,14 @@ config :mongoose_push, fcm_enabled: {:system, :boolean, "PUSH_FCM_ENABLED", true
...
+  tls_server_cert_validation: {:system, :boolean, "TLS_SERVER_CERT_VALIDATION", true}
+
 config :mongoose_push,
   fcm: [
     default: [
       endpoint: {:system, :string, "PUSH_FCM_ENDPOINT", nil},
+      port: {:system, :integer, "PUSH_FCM_PORT", nil},
```
rslota, line 33: Don't we need something similar for APNS?

integration.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+docker network create n
+
+docker run -p 4000:4000 -p 4001:4001 --name fcm --network n mongooseim/fcm-mock-server
+
+docker run -v `pwd`/priv:/opt/app/priv -e PUSH_FCM_ENABLED=true -e PUSH_FCM_PORT=4000 -e PUSH_FCM_ENDPOINT="fcm" -e PUSH_APNS_ENABLED=false -e TLS_SERVER_CERT_VALIDATION=false -e FCM_AUTH_ENDPOINT="http://fcm:4001" --network n -p 8443:8443 mongooseim/mongoose-push:zp_integration-tests
+
+MIX_ENV=integration mix test test/api --only integration
```
rslota: I know that this is only temporary file, but in final version this line should be implied by `MIX_ENV=integration`. Tests path, test tag and disabled app startup should all be defined in `mix.exs` for this env.

.circleci/config.yml
```diff
@@ -4,6 +4,21 @@ filters: &all_tags
   tags:
     only: /.*/
 
+run: &DOCKER_TAGS
+    name: Generate image tag
+    command: |
+        set -x
+        if [ -n "$CIRCLE_TAG" ]; then
+        DOCKER_TAG="$CIRCLE_TAG"
```
rslota: Would be nice to add some indentations in this code block.

.circleci/config.yml
```diff
@@ -175,64 +160,129 @@ jobs:
...
+          command: sudo dockerize -wait tcp://localhost:4000 -timeout 1m
+      - run:
+          name: Wait for APNS mock
+          command: sudo dockerize -wait tcp://localhost:2197 -timeout 1m
+      - run:
+          name: Wait for MongoosePush
+          command: sudo dockerize -wait tcp://localhost:8443 -timeout 2m
+      - run:
+          name: Run integration tests
+          command: MIX_ENV=integration mix do deps.get, deps.compile, test test/api --only integration
```
rslota: This like should simply be:
```
 MIX_ENV=integration mix do deps.get, test
```
Everything else should be set by `mix.exs`.

integration.sh
rslota: Do we keep this file? I see very little gain from it, and even if the is some, it's better to write docker-compose file to do this.

test/api/v3_apns_test.exs
```diff
@@ -1,15 +1,29 @@
 defmodule MongoosePush.API.V3APNSTest do
   use ExUnit.Case, async: false
   alias MongoosePush.Support.API, as: Tools
-  doctest MongoosePush.API.V3
 
   @url "/v3/notification/f534534543"
 
+  setup_all do
+    if Mix.env() == :integration do
+      HTTPoison.start()
```
rslota: Why isn't it starting? This seems hacky - it should be there already, and if it's not, then something may be wrong and needs fixing.

test/api/v3_apns_test.exs
```diff
@@ -1,15 +1,29 @@
...
+
+    :ok
+  end
+
   setup do
-    Tools.reset(:apns)
-    TestHelper.reload_app()
+    case Mix.env() do
+      :test ->
+        Tools.reset(:apns)
```
rslota: Mock reset is for both cases, so can be put outside of this condition.

test/api/v3_fcm_test.exs
```diff
@@ -1,15 +1,29 @@
 defmodule MongoosePush.API.V3FCMTest do
   use ExUnit.Case, async: false
   alias MongoosePush.Support.API, as: Tools
-  doctest MongoosePush.API.V3
 
   @url "/v3/notification/f534534543"
 
+  setup_all do
+    if Mix.env() == :integration do
+      HTTPoison.start()
```
rslota: Why isn't it starting? This seems hacky - it should be there already, and if it's not, then something may be wrong and needs fixing.

test/api/v3_fcm_test.exs
```diff
@@ -1,15 +1,29 @@
...
+
+    :ok
+  end
+
   setup do
-    Tools.reset(:fcm)
-    TestHelper.reload_app()
+    case Mix.env() do
+      :test ->
+        Tools.reset(:fcm)
```
rslota: Mock reset is for both cases, so can be put outside of this condition.

.circleci/config.yml
```diff
@@ -175,64 +160,129 @@ jobs:
...
+              - certs-cache
+      - run: *DOCKER_TAGS
+      - run:
+          name: Setup mocks and MPush
+          command: |
+              docker network create integration || true && \
+              docker run -d -p 4000:4000 -p 4001:4001 --name fcm --network integration mongooseim/fcm-mock-server && \
+              docker run -d -p 2197:2197 --name apns --network integration mobify/apns-http2-mock-server && \
+              echo "echo $PWD/app/priv" >> $BASH_ENV && \
+              docker run -d -v /home/circleci/app/priv:/opt/app/priv \
```
rslota: How about creating a docker-compose file that would setup all this? It would also be usable for local environment and could reside next to integration tests. The tests do/may assume some things about this mpush configuration, so it actually make more sense to have all this within MongoosePush codebase, not CI configuration. I think the only variables are the docker image name / tag, which can be taken from system env anyway (docker-compose does support such case).

rslota (changes requested):

config/prod.exs
zofpolkowska: We already have `use_2197` option there.

config/prod.exs
zofpolkowska: Done

test/integration/connection_lost_test.exs
```diff
@@ -0,0 +1,27 @@
+defmodule MongoosePush.API.V3FCMTest do
+  use ExUnit.Case, async: false
+  use HelperMacros
+  alias MongoosePush.Support.API, as: Tools
+
+  @url "/v3/notification/f534534543"
+
+  setup do
+    if Mix.env() == :test do
```
rslota: This will never happen, it's a dead code.

test/integration/test_helper.exs
```diff
@@ -2,5 +2,35 @@ ExUnit.start(capture_log: true)
 
 HTTPoison.start()
 
+defmodule HelperMacros do
```
rslota: We should move this module to `support` so that we don't redefine. Also, https://github.com/rslota/ex_assert_eventually <- much better version available as dependency.

.circleci/config.yml
rslota: Bump :)

README.md
```diff
@@ -72,6 +72,8 @@ Environmental variables to configure production release:
 * `PUSH_LOGLEVEL` - `debug`/`info`/`warn`/`error` - Log level of the application. `info` is the default one
 * `PUSH_FCM_ENABLED` - `true`/`false` - Enable or disable `FCM` support. Enabled by default
 * `PUSH_APNS_ENABLED` - `true`/`false` - Enable or disable `APNS` support. Enabled by default
+* `TLS_SERVER_CERT_VALIDATION` - `true`/`false` - Enable or distable TLS
```
rslota: Note about prod env is redundant - only this is being shipped and is configurable with envs.

mix.exs
```diff
@@ -101,10 +104,16 @@ defmodule MongoosePush.Mixfile do
 
   # All mix tasks are redundant in runtime, but we still need to compile `lib/mix/tasks/compile_asn1.ex`
   # as it's required by build process (ASN1 compiler).
-  defp elixirc_paths(:prod),
-    do: ["lib"]
 
-  defp elixirc_paths(:test), do: ["lib", "test/support"]
+  defp elixirc_paths(:prod), do: ["lib"]
+
+  defp elixirc_paths(:test), do: ["lib", "test/support", "test/common", "test/unit"]
```
rslota: There is not reason for this condition - tests do not compile, thus there is no reason to setup different compile paths.

test/common/test_helper.exs
```diff
@@ -0,0 +1,34 @@
+ExUnit.start(capture_log: true)
+
+if Mix.env() == :integration do
```
rslota: Why not start it each time?

test/common/test_helper.exs
```diff
@@ -0,0 +1,34 @@
...
+  HTTPoison.start()
+end
+
+defmodule TimeHelper do
```
rslota: Why do we even have common test_helper? test helper is supposed to bootstrap tests. Code should go to `support` to be compiled.

rslota (changes requested):

test/common/test_helper.exs
zofpolkowska: Don't we need a `test_helper` in all of the `test_paths`?

rslota: I don't think so, but I may be mistaken.

test/integration/test_helper.exs
```diff
@@ -0,0 +1,6 @@
+ExUnit.start(capture_log: true)
+
+HTTPoison.start()
+
+defmodule TestHelper do
```
rslota: Why do we needs this here?

test/docker/integration.yml
```diff
@@ -0,0 +1,34 @@
+version: '3'
+
+services:
+    fcm-mock:
+        image: mongooseim/fcm-mock-server
+        container_name:  fcm-mock
+        ports:
+            - "4001:4001"
```
rslota, line 8: I don't remember which is HTTPS, but the HTTPS is not needed here, as pretty much only mpush should connect to it - and since mpush is within docker-compose network, binding locally is superfluous.

test/docker/integration.yml
```diff
@@ -0,0 +1,34 @@
...
+            - PUSH_APNS_DEV_USE_2197=true
+            - PUSH_FCM_ENABLED=true
+            - PUSH_FCM_PORT=4000
+            - PUSH_FCM_ENDPOINT=fcm-mock
+            - TLS_SERVER_CERT_VALIDATION=false
+            - FCM_AUTH_ENDPOINT=http://fcm-mock:4001
+        ports:
+            - "8443:8443"
+        volumes:
+            - /home/circleci/app/priv:/opt/app/priv
```
rslota: I would rather not make this CircleCI dependent. Ether let's get the path from ENV, or assume that compose will be used from main directory of the project in which case path `$(PWD)/priv` should be valid and equivalent, but more generic.

test/common/test_helper.exs
```diff
@@ -1,20 +1,6 @@
...
-    rescue
-      ExUnit.AssertionError ->
-        :timer.sleep(10)
-        wait_until(max(0, timeout - 10), fun)
-    end
-  end
-end
+HTTPoison.start()
 
 defmodule TestHelper do
```
rslota: This still seems to be duplicated (saw that in same other file).

rslota (changes requested):

test/integration/test_helper.exs
zofpolkowska: Why do we need what?

test/docker/integration.yml
zofpolkowska: Well this is confusing in cci

test/common/test_helper.exs
zofpolkowska: Removed from other place now

test/docker/integration.yml
zofpolkowska: Apparently it fails to connect if we do not expose both

mix.exs
```diff
@@ -101,10 +105,17 @@ defmodule MongoosePush.Mixfile do
...
+
+  defp elixirc_paths(:prod), do: ["lib"]
 
   defp elixirc_paths(:test), do: ["lib", "test/support"]
 
+  defp elixirc_paths(:integration), do: ["test/support"]
+
   defp elixirc_paths(_), do: ["lib"]
+
+  # Add "test/integration" path after the https://erlangsolutions.atlassian.net/browse/MIM-748 is resolved
```
rslota: Let's configure this file properly and use `@skip` tag instead to skip not-working tests.

test/docker/integration.yml
```diff
@@ -0,0 +1,33 @@
...
+            - PUSH_APNS_DEV_USE_2197=true
+            - PUSH_FCM_ENABLED=true
+            - PUSH_FCM_PORT=4000
+            - PUSH_FCM_ENDPOINT=fcm-mock
+            - TLS_SERVER_CERT_VALIDATION=false
+            - FCM_AUTH_ENDPOINT=http://fcm-mock:4001
+        ports:
+            - "8443:8443"
+        volumes:
+            - ${HOME}/app/priv:/opt/app/priv
```
rslota: Why home? This will also not work anywhere outside of CI, as this requires to have `/app/priv` in $HOME. Why not $PWD? If for some reason $PWD doesn't work, let's go with custom env. The point of having compose here is that we can use it on any CI and locally.

test/integration/test_helper.exs
rslota: This module. It's empty anyway.

rslota: Not to mention, that putting it here was a mistake from beginning - scripts should never declare modules, unless they itself need it.

rslota (changes requested):

test/docker/integration.yml
rslota: Why is it confusing? CCI allows to configure working dir, so everything can be explicit. We can configure work dir to X, clone to X, `cd X` and then PWD is exacly what we want when running compose.

.circleci/config.yml
```diff
@@ -175,64 +160,113 @@ jobs:
...
+              sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-v0.6.1.tar.gz && \
+              sudo chmod +x /usr/local/bin/dockerize && \
+              sudo rm dockerize-linux-amd64-v0.6.1.tar.gz
+
+      - restore_cache:
+          keys:
+              - mix-cache-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}
+      - restore_cache:
+          keys:
+              - certs-cache
```
rslota: We really should make this branch specific and add `{{ .Environment.CACHE_VERSION }}` because we may have conflicts between branches because of that.

rslota (approved):

PR #134, opened 2020-01-27
Add integration test for lost connection

kmakiela:


test/integration/connection_lost_test.exs
```diff
@@ -11,15 +11,30 @@ defmodule MongoosePush.API.ConnectionTest do
...
+  def try_post(url, notification, retry_no) do
+    if retry_no == 0 do
+      :not_ok
+    else
+      case Tools.post_conn_error(url, notification) do
+        {200, _} ->
+          :ok
+
+        sth ->
+          IO.inspect(sth)
```
rslota: For removal :)

test/integration/connection_lost_test.exs
```diff
@@ -11,15 +11,30 @@ defmodule MongoosePush.API.ConnectionTest do
     :ok
   end
 
-  # We have to skip it until MIM-748 is resolved
-  @tag :skip
   test "When connection to FCM is lost and regained a PN succeeds" do
     Tools.mock_fcm("/connection", %{"https" => false})
 
-    eventually(assert {500, ""} == Tools.post_conn_error(@url, Tools.sample_notification(:fcm)))
+    assert :not_ok == try_post(@url, Tools.sample_notification(:fcm), 5)
```
rslota: We actually need to make sure the error is correct here (should be HTTP 503, another bug in this test).
How about:
```elixir
Enum.each(1..5, fn _ -> 
  eventually(assert {503, ""} == Tools.post_conn_error(@url, Tools.sample_notification(:fcm)))
  Process.sleep(1000)
end)
```

rslota (changes requested):

test/integration/connection_lost_test.exs
```diff
@@ -11,15 +11,37 @@ defmodule MongoosePush.API.ConnectionTest do
...
 
-  # We have to skip it until MIM-748 is resolved
-  @tag :skip
   test "When connection to FCM is lost and regained a PN succeeds" do
     Tools.mock_fcm("/connection", %{"https" => false})
 
-    eventually(assert {500, ""} == Tools.post_conn_error(@url, Tools.sample_notification(:fcm)))
+    # assert :not_ok == try_post(@url, Tools.sample_notification(:fcm), 5)
+    Enum.each(1..5, fn _ ->
+      eventually(assert {500, ""} == Tools.post_conn_error(@url, Tools.sample_notification(:fcm)))
```
rslota: As I've mentioned - this should be 503 error.

test/integration/connection_lost_test.exs
```diff
@@ -11,15 +11,37 @@ defmodule MongoosePush.API.ConnectionTest do
...
 
-    eventually(assert {200, nil} == Tools.post(@url, Tools.sample_notification(:fcm)))
+    # assert :ok == try_post(@url, Tools.sample_notification(:fcm), 5)
+    Enum.each(1..5, fn _ ->
+      eventually(assert {200, _} == Tools.post_conn_error(@url, Tools.sample_notification(:fcm)))
+      Process.sleep(1000)
+    end)
+  end
+
+  def try_post(url, notification, retry_no) do
```
rslota: This doesn't seem to be needed anymore.

rslota (changes requested):

test/integration/connection_lost_test.exs
kmakiela: It seems there is still code 500 returned

rslota: So we need to change that :)

kmakiela: The reason for that is probably because Sparrow returns unknown error - current MP version does not include the Sparrow version that unlinks the worker

rslota: It should - this PR should enable the test, update sparrow and bump the version.

rslota (approved):

PR #135, opened 2020-02-12
Add telemetry support

kmakiela:
* Added telemetry calls

* Added Telemetry behaviour

* Added handlers for APNS and FCM events

* Moved Elixometer calls to telemetry events


lib/mongoose_push.ex
```diff
@@ -108,4 +110,27 @@ defmodule MongoosePush do
     Logger.warn(~s"Unable to complete push request due to #{inspect(reason)}")
     return_value
   end
+
+  defp emit_telemetry_event(time, {:error, {type, reason}}) do
+    :telemetry.execute(
+      [:mongoose_push, :push, :error, type],
```
rslota: Event name should be constant. `type` of error can be a label (metdata).

lib/mongoose_push.ex
```diff
@@ -108,4 +110,27 @@ defmodule MongoosePush do
...
+    :telemetry.execute(
+      [:mongoose_push, :push, :error],
+      %{time: time},
+      %{reason: reason}
+    )
+  end
+
+  defp emit_telemetry_event(time, :ok) do
+    :telemetry.execute(
+      [:mongoose_push, :push, :ok],
```
rslota:
```suggestion
      [:mongoose_push, :push, :success],
```

lib/mongoose_push/service/apns/state.ex
```diff
@@ -15,6 +15,15 @@ defmodule MongoosePush.Service.APNS.State do
     default_topics = setup_default_topics(pool_configs)
     :apns_state = :ets.new(:apns_state, [:set, :public, :named_table])
     true = :ets.insert(:apns_state, default_topics)
+
+    :telemetry.execute(
+      [:mongoose_push, :apns_state, :init],
```
rslota: `init` events should always get their `terminate` counterpart in order to enable metrics like "number of active instances".

lib/mongoose_push/service/apns/supervisor.ex
```diff
@@ -32,6 +32,12 @@ defmodule MongoosePush.Service.APNS.Supervisor do
           {MongoosePush.Service.APNS.State, pool_configs}
         ]
 
+        :telemetry.execute(
+          [:mongoose_push, :apns_supervisor, :init],
```
rslota: `init` events should always get their `terminate` counterpart in order to enable metrics like "number of active instances".

lib/mongoose_push/service/fcm/pool/supervisor.ex
```diff
@@ -24,6 +24,12 @@ defmodule MongoosePush.Service.FCM.Pool.Supervisor do
       )
     ]
 
+    :telemetry.execute(
+      [:mongoose_push, :fcm_supervisor, :init],
```
rslota: `init` events should always get their `terminate` counterpart in order to enable metrics like "number of active instances".

lib/mongoose_push/telemetry/handler.ex
```diff
@@ -0,0 +1,42 @@
+defmodule MongoosePush.Telemetry.Handler do
+  @moduledoc """
+  Module responsible for handling telemetry events, in future should be integrated with Prometheus
```
rslota: Maybe let's not use `Prometheus` name explicitly, as we haven't officially decided that, and also, this doesn't actually matter.

lib/mongoose_push/telemetry/handler.ex
```diff
@@ -0,0 +1,42 @@
...
+  """
+  use GenServer
+  require Logger
+
+  def start_link(_) do
```
rslota: Is there any reason this module is a `GenServer`? This seems a bit too much considering it's use case is very simple, while adding async process here adds a lot complexity regarding back pressure issues (more events then target metric engine can handle), or simply slowing down event processing due to centralized nature of this `GenServer`.

lib/mongoose_push.ex
```diff
@@ -86,6 +86,8 @@ defmodule MongoosePush do
         :timer.tc(module, :push, [notification, device_id, pool, opts])
       end
 
+    emit_telemetry_event(time, push_result)
+
     push_result
     |> Metrics.update(:spiral, [:push, service, mode])
```
rslota: Metrics updates in elixometer should not longer be visible anywhere in core logic. Those calls should be made in telemetry handlers.

lib/mongoose_push/telemetry.ex
```diff
@@ -0,0 +1,31 @@
+defmodule MongoosePush.Telemetry do
+  @moduledoc """
+  Module responsible for handling telemetry events
+  """
+  alias MongoosePush.Telemetry.Handler
+
+  def handler do
+    attach_metrics()
```
rslota: This feels weird. We attach handlers each time we call function that returns supervisor spec? This seems like side effect that is hard to expect. I would much prefer this to be just a separate function, like. `attach_all/0` that is meant to be explicitly called when application starts.

lib/mongoose_push/telemetry.ex
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  defdelegate handle_event(tags, measurements, metadata, config), to: Handler, as: :handle_event
+
+  def attach_metrics do
+    _res =
+      :telemetry.attach_many(
+        "handler",
+        [
+          [:mongoose_push, :push],
```
rslota: I feel that we should categorize the events more, maybe have separate handler modules for each service + core.
For this to happen, we need to change some event names, to always start off with one of:
* `[:mongoose_push, :apns, ....]`
* `[:mongoose_push, :fcm, ....]`
* `[:mongoose_push, :core, ....]`

rslota: When we start handling sparrow events, we'll need to add more handler modules anyway, so that they are not too big and well scoped, so we may as well start here.

rslota (changes requested):
In general this is the direction we want to go, but as part of this PR, elixometer should be moved to be a handler for telemetry events, so that MongoosePush code itself only deals with telemetry.

lib/mongoose_push/service/apns/supervisor.ex
kmakiela: Here we use `Supervisor` and not `GenServer` and it doesn't have a `terminate` callback - not sure how to handle that

lib/mongoose_push/service/fcm/pool/supervisor.ex
kmakiela: Here we use `Supervisor` and not `GenServer` and it doesn't have a `terminate` callback - not sure how to handle that

lib/mongoose_push/telemetry/handler.ex
kmakiela: As long as we just update some metrics I agree that it's not needed.

lib/mongoose_push/telemetry.ex
```diff
@@ -0,0 +1,36 @@
+defmodule MongoosePush.Telemetry do
+  @moduledoc """
+  Module responsible for handling telemetry events
+  """
+
+  def attach_all do
+    _ =
+      :telemetry.attach_many(
```
rslota: How about splitting this into two functions so that it scales better when we add more?

rslota: Actually, how about chaning this module into behaviour that `MongoosePush.Telemetry.APNSHandler` and `MongoosePush.Telemetry.FCMHandler` implement, with two functions: `handle_event/4` and `attach/0`? This way each handler module registers its own handlers.

rslota (changes requested):
Looks great, though I do have one small code style suggestion. Also, there is one metric that hasn't been moved to being telemetry handler - the main `push` function is tagged with `@timed key: :auto`. This also should be achievable with just telemetry and thus moved to handlers in form of direct metric update.

rslota (approved):

PR #139, opened 2020-02-17
Change logs to be structured

kmakiela:


michalwski:
Does Elixir allow us to produce structured logs as described in https://ferd.ca/erlang-otp-21-s-new-logger.html by passing a map and letting the formatter printing as key=value pairs?

rslota:
@michalwski @kmakiela ,

> Does Elixir allow us to produce structured logs as described in https://ferd.ca/erlang-otp-21-s-new-logger.html by passing a map and letting the formatter printing as key=value pairs?

Yes, it does. Elixir's logger can easily get a map as a second argument but string text as first has to also be there. Then with custom formatter we can do whatever we feel like doing. 
I strongly suggest going this way, and implementing `logfmt` formatter as first one, and hopefully `json` one as second. `logfmt` could be the default one, as the one that allows for ordering of keys and that is more human-readable, but `json` is much, much easier to work with is logs are being indexed by some external log collectors. 

I honestly wanted to point you to `sparrow`, but now I can see that `sparrow` actually also uses string interpolation instead, so also needs a rewrite. Using string-interpolation does not allow us to control which data is printed depending on context (`logfmt` should print minimal useful set, white `json` as it's only meant to be indexed, could print all data available) and allows for only one format.

config/config.exs
```diff
@@ -7,8 +7,8 @@ config :logger,
   handle_sasl_reports: false
 
 config :logger, :console,
-  format: "\n$dateT$time [$level] $metadata$levelpad$message\n",
-  metadata: [:pid]
+  format: {MongoosePush.Logger, :format},
```
rslota: I would strongly suggest to call this module based on the actual format it produces, as there may be many (and probably will).

lib/mongoose_push/logger.ex
```diff
@@ -0,0 +1,25 @@
+defmodule MongoosePush.Logger do
+  @moduledoc """
+  Module responsible for formatting logger output
+  """
+  def format_fmt(level, message, {date, time}, metadata) do
+    date_f = Logger.Formatter.format_date(date)
+    time_f = Logger.Formatter.format_time(time)
+    meta_f = Logfmt.encode(metadata)
+
+    "\n#{date_f} #{time_f} [#{level}] #{meta_f} #{message}"
```
rslota: The whole message should be encoded in the same format, there is very little use if only part of it is standardized.

rslota: Hmm, although for logfmt specifically we may do this differently to make it more readable - but this exception I would apply at most for date and log level - message should be metadata.

config/config.exs
kmakiela: I thought both `fmt` and `json` will have dedicated functions inside the same module - do we want to have a module for each logger formatter?

rslota: Yeah, event if that's not required (isn't formatter a behavior?), I would vote for them to be different modules.

lib/mongoose_push/logger.ex
kmakiela: What exactly do we want to pass as a message then? Do we want it to have a specific format or should it just be a text description?

rslota: Message should just be part of `metadata` map under key `msg` or something like that.
The only special treatment it can get is to position it "better". Please remember that `logfmt` does have positional formatting, unlike `json`.

kmakiela: My doubts were about what to put as a message and what as metadata when calling `Logger.log` - the logger docs mentioned above have fields like `what` , `at` and `result` which cover the log message that existed before completely.

rslota: I think in this naming, `what` == `msg`, at least I think so. Regardless of the name of this field - Elixir's Logger requires the 1st argument to be a string, and thus we have to put the message there. Everything else as metadata. On formatter level we can however merge those two and do whatever we feel like it.

config/config.exs
```diff
@@ -7,8 +7,8 @@ config :logger,
   handle_sasl_reports: false
 
 config :logger, :console,
-  format: "\n$dateT$time [$level] $metadata$levelpad$message\n",
-  metadata: [:pid]
+  format: {MongoosePush.LoggerFmt, :format},
```
rslota, line 10: Let's go with more structured approach - `MongoosePush.Logger.Logfmt` (App.Module.Format).

lib/mongoose_push.ex
```diff
@@ -72,7 +72,16 @@ defmodule MongoosePush do
 
     {time, push_result} =
       if pool == nil do
-        Logger.error(~s"No pool matching mode=#{mode} and tags=#{inspect(tags)}")
+        Logger.error(
+          fn ->
```
rslota, line 76: Do we really need this function? This is error log, it's very unusual for them to compiled-out.

lib/mongoose_push.ex
```diff
@@ -72,7 +72,16 @@ defmodule MongoosePush do
...
+            "choose_pool"
```
rslota, line 77: This message should be "human readable", yet without any variables. Here I would suggest - "No matching pool found."

lib/mongoose_push.ex
```diff
@@ -96,16 +105,21 @@ defmodule MongoosePush do
...
   defp maybe_log({:error, {type, reason}} = return_value) do
     Logger.warn(
-      ~s"Unable to complete push request due to service error: #{inspect(reason)} in category: #{
-        type
-      }"
+      fn ->
+        "push"
+      end,
+      result: :error,
+      log: :command,
```
rslota, line 112: What is the meaning of this label?

lib/mongoose_push.ex
```diff
@@ -96,16 +105,21 @@ defmodule MongoosePush do
...
+      reason: reason,
+      type: type
     )
 
     return_value
   end
 
   defp maybe_log({:error, reason} = return_value) do
-    Logger.warn(~s"Unable to complete push request due to #{inspect(reason)}")
+    Logger.warn(fn -> "push" end, result: :error, log: :command, reason: reason)
```
rslota, line 121: For message I would suggest - "Unable to complete push request". In general messages should be "human readable", but should not contain any variables. It's actually very similar to events name in telemetry, only stringified.

lib/mongoose_push/logger_fmt.ex
```diff
@@ -0,0 +1,25 @@
...
+    line = Keyword.fetch!(metadata, :line)
+    pid = Keyword.fetch!(metadata, :pid)
+
+    meta_f =
+      [what: message]
+      |> Keyword.put(:at, "#{module}.#{function}:#{line}")
+      |> Keyword.put(:pid, pid)
+      |> Logfmt.encode()
+
+    "\n#{date_f} #{time_f} [#{level}] #{meta_f}"
```
rslota, line 21: I don't seem to find other `metadata` that we pass.

rslota (changes requested):
Few comments, but those apply to pretty much all logs ;)

PR #144, opened 2020-03-20
MIM-864 mongoosepush exposes phoenix initial endpoint

leszke:
From story description:

As MongoosePush Client (HTTP API user)

Given MongoosePush is running

When I send a dummy POST request to MongoosePush on Phoenix backed endpoint

Then I get dummy response

 

Notes:

- Both HTTP and HTTPS protocols should be supported 

- Both HTTP 1.1 and HTTP 2.0 protocols should be supported 

- Temporary we need to expose those two endpoints on different ports then current API

- Phoenix should be configured with no Ecto and no Views. 

- Phoenix Websockets should be disabled

- No need for exposing new endpoints in release Docker container

- Dummy Controller and dummy entry in Router should be created in order to propose initial file structure - we need to keep in mind protocol versions split etc.


config/dev.exs
```diff
@@ -61,3 +61,16 @@ config :mongoose_push,
...
+  https: [
+    port: 8446,
+    keyfile: "priv/ssl/fake_key.pem",
+    certfile: "priv/ssl/fake_cert.pem",
+    otp_app: :mongoose_push
+  ],
+  debug_errors: true,
+  code_reloader: false,
+  check_origin: false,
+  server: true
```
rslota: Missing end-of-line.

lib/mongoose_push/application.ex
```diff
@@ -32,7 +32,7 @@ defmodule MongoosePush.Application do
     MongoosePush.Telemetry.attach_all()
 
     # Define workers and child supervisors to be supervised
-    children = children()
+    children = children() ++ [MongoosePushWeb.Endpoint]
```
rslota: Maybe let's rename `children` to `service_children` or something like that in order to indicate that those are not "all possible children" anymore.

lib/mongoose_push_web/controllers/dummy_controller.ex
```diff
@@ -0,0 +1,7 @@
+defmodule MongoosePushWeb.DummyController do
```
rslota, line 1: This file uses 4 spaces indentation, while other use double. We should normalize to double-spaces.

lib/mongoose_push_web/controllers/dummy_controller.ex
```diff
@@ -0,0 +1,7 @@
...
+    use MongoosePushWeb, :controller
+
+    def handle(conn, params) do
+        json(conn, %{body: params})
+    end
+end
```
rslota: Missing end-of-line character.

lib/mongoose_push_web/endpoint.ex
```diff
@@ -0,0 +1,41 @@
...
+    store: :cookie,
+    key: "_mongoose_push_key",
+    signing_salt: "URIBGSwL"
+  ]
+
+  # Serve at "/" the static files from "priv/static" directory.
+  #
+  # You should set gzip to true if you are running phx.digest
+  # when deploying your static files in production.
+  plug Plug.Static,
```
rslota: I don't think we need any static data, so we can remove this plug.

test/mongoose_push_web/controllers/dummy_controller_test.exs
```diff
@@ -0,0 +1,13 @@
+defmodule MongoosePushWeb.DummyControllerTest do
```
rslota, line 1: This file uses 4 spaces indentation, while other use double. We should normalize to double-spaces.

test/mongoose_push_web/controllers/dummy_controller_test.exs
```diff
@@ -0,0 +1,13 @@
...
+    test "POST /api/dummy", %{conn: conn} do
+        conn = post(conn, "/api/dummy")
+        assert json_response(conn,200) == %{"body" => %{}}
+    end
+
+    test "POST /api/dummy with parameters", %{conn: conn} do
+        conn = post(conn, "/api/dummy", [key: "value"])
+        assert json_response(conn,200) == %{"body" => %{"key"=>"value"}}
+      end
+  end
```
rslota: Missing end-of-line character.

test/mongoose_push_web/test_helper.exs
```diff
@@ -0,0 +1,2 @@
+{:ok, _} = Application.ensure_all_started(:mongoose_push)
+ExUnit.start()
```
rslota: Missing end-of-line character.

rslota: Also, it's recommended to use `ExUnit.start(capture: true)` in order to hide logs as long tests are green.

lib/mongoose_push_web/endpoint.ex
```diff
@@ -0,0 +1,41 @@
...
+  use Phoenix.Endpoint, otp_app: :mongoose_push
+
+  # The session will be stored in the cookie and signed,
+  # this means its contents can be read but not tampered with.
+  # Set :encryption_salt if you would also like to encrypt it.
+  @session_options [
+    store: :cookie,
+    key: "_mongoose_push_key",
+    signing_salt: "URIBGSwL"
+  ]
```
michalwski: From what I understand, we don't need this in MongoosePush since there is no session needed.

lib/mongoose_push_web/endpoint.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  # Serve at "/" the static files from "priv/static" directory.
+  #
+  # You should set gzip to true if you are running phx.digest
+  # when deploying your static files in production.
+  plug Plug.Static,
+    at: "/",
+    from: :mongoose_push,
+    gzip: false,
+    only: ~w(css fonts images js favicon.ico robots.txt)
```
michalwski: This one is also not needed. We'll probably don't have static files to serve.

rslota (changes requested):
Thanks for the PR! In general, the code does what it suppose to, just few minor things to change, mostly formatting. I'm not sure if `mix format` has been used here, if not, let's do so.
Quite a few files are missing newline character at the end of last line - if that was intentional, I recommend this read: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file :)

lib/mongoose_push_web/endpoint.ex
```diff
@@ -0,0 +1,41 @@
...
+    plug Phoenix.CodeReloader
+  end
+
+  plug Plug.RequestId
+  plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
+
+  plug Plug.Parsers,
+    parsers: [:urlencoded, :multipart, :json],
+    pass: ["*/*"],
+    json_decoder: Phoenix.json_library()
```
michalwski, line 16: Do we know what is the default json encoder? I don't see it configured and I think it's better to use the one which is already used in the project. As far as I know MongoosePush already uses Jason. It can be configured by adding the following line:
```elixir
config :phoenix, :json_library, Jason
```
to `config/config.exs`

michalwski (commented):
Thanks for the changes @leszke! I left some comments to parts of the code which I think is not required by MongoosePush and a question regarding Phoenix's JSON library.

lib/mongoose_push_web/endpoint.ex
michalwski: @rslota what do you think?

lib/mongoose_push_web/endpoint.ex
michalwski: @rslota what do you think?

lib/mongoose_push_web/endpoint.ex
rslota: Yeah, it seems that we can remove this along with entire `Plug.Session`.

lib/mongoose_push_web/endpoint.ex
rslota: Yeah, also left comment for that :)

config/dev.exs
leszke: It was unintentional and corrected with `mix format`. Anyway thank you for comprehensive reading :)

michalwski (approved):

PR #145, opened 2020-03-25
[WIP] MIM-865 Define APIv1 OpenAPI specification

leszke:
From story description:

As a MongoosePush Client

Given MongoosePush is running

When I send a GET /swagger.json HTTP request to Phoenix backed endpoint

Then I get JSON formatted OpenAPI 3.0 specification for MongoosePush APIv1

Notes:

- The easiest and best way of writing OpenAPI 3.0 spec in Elixir is using: https://github.com/open-api-spex/open_api_spex

- This story requires defining APIv1 from scratch using the mentioned library based on current documentation and/or Swagger 2.0 spec.

- The JSON formatted OpenAPI 3.0 specification can be exposed using Plug included in the mentioned library.


lib/mongoose_push_web/controllers/api_v1_controller.ex
```diff
@@ -0,0 +1,34 @@
+defmodule MongoosePushWeb.APIv1Controller do
+  alias MongoosePushWeb.Schemas
+  alias OpenApiSpex.Operation
+  use MongoosePushWeb, :controller
+
+  @spec open_api_operation(atom) :: Operation.t()
+  def open_api_operation(action) do
```
rslota: We're gonna have 3 APIs, 3 controllers, yet this function will always be the same. Also, the dynamic function apply is not "pretty", yet required, so hiding this would IMHO be helpful. How about injecting this function with `using` macro that could come with `MongoosePushWeb.Schemas` module? We would endup with:

```
use MongoosePushWeb.Schemas
```

lib/mongoose_push_web/controllers/api_v1_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+    apply(__MODULE__, operation, [])
+  end
+
+  @spec handle_operation() :: Operation.t()
+  def handle_operation() do
+    %Operation{
+      tags: ["apiv1"],
+      summary: "sends a push",
+      description: "performs the sending of push notification",
+      operationId: "APIv1Controller.handle",
```
rslota: This operation should not be called `handle`. In fact, no operations should be called that. This is not meaningful. Let's imagine we add second thing to the API, e.g. `/api/v1/batch_push`, how would it be called? 
IMHO, this one should be simply `push`.

lib/mongoose_push_web/controllers/api_v1_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+      parameters: [
```
rslota: I'm pretty sure we don't need that - more research needed.

lib/mongoose_push_web/controllers/api_v1_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+        Operation.parameter(:id, :path, :string, "Device ID", example: "f53453455", required: true)
+      ],
+      requestBody:
+        Operation.request_body(
+          "The push notification attributes",
+          "application/json",
+          Schemas.APIv1Request,
```
rslota: Same thing as with `handle` operation - this is too generic name. This is `push` request for APIv1. Also, let's try to split the naming into more submodules, like `Schemas.APIv1.PushRequest`, or `Schemas.APIv1.Request.Push`.

rslota: We need to figure out the structure here, as some input / output objects will be shared between API versions, and some will be completely different (APIv2 and APIv3 share entire request object, those differ only for response).

lib/mongoose_push_web/controllers/api_v1_controller.ex
rslota: Or maybe `send_notification`?

lib/mongoose_push_web/router.ex
```diff
@@ -5,9 +5,26 @@ defmodule MongoosePushWeb.Router do
...
+  scope "/" do
+    pipe_through(:swagger_json)
+
+    get("/swagger.json", OpenApiSpex.Plug.RenderSpec, [])
+  end
+
+  scope "/v1", MongoosePushWeb do
+    pipe_through(:api)
+
+    post("/notification", APIv1Controller, :handle)
```
rslota: How about:

```suggestion
    post("/notification", APIv1.PushController, :send)
```

rslota: Controllers always should group operations one the same objects (notification), while handlers should represent operations on this object (send?).

lib/mongoose_push_web/schemas.ex
```diff
@@ -0,0 +1,36 @@
+defmodule MongoosePushWeb.Schemas do
+  alias OpenApiSpex.Schema
+
+  defmodule APIv1Request do
```
rslota: Let's never do that :)
We should always have 1-to-1 mapping with modules <-> files.

lib/mongoose_push_web/schemas.ex
```diff
@@ -0,0 +1,36 @@
...
+    require OpenApiSpex
+
+    OpenApiSpex.schema(%{
+      title: "User",
+      description: "A user of the app",
+      type: :object,
+      properties: %{
+        service: %Schema{type: :string, description: "Push notification service", format: :string},
```
rslota: I'm pretty sure we can define enum-type here somehow to list legal values.

rslota (changes requested):
Thanks for the PR! The general use of OpenApiSpec seems to be just fine, but I do have some comments on general file/module placement/naming.

lib/mongoose_push_web/controllers/api_v1_controller.ex
michalwski: I'm not sure parameters in the request's path can be described in a different way. At least, I didn't find it defined differently in the lib's doc.

michalwski: From my understanding, the schemas can be used to define the request body. Other parameters like these in `path`, `query`, `header` or `cookie` are described with thise API: https://hexdocs.pm/open_api_spex/OpenApiSpex.Operation.html#parameter/5

michalwski: Sorry, I just saw it :) it is possible to pass the schema reference to the `Operation.parameter` function as 3rd argument. The question is do we want that? On one hand, this probably common for all versions of the IP, on the other hand the type is just a string.

rslota: No, no, why do we even need parameters? Like you said - those are to define `path`, `query`, `header` or `cookie`, none of which MongoosePush uses.

rslota: `parameters` is just a HTTP's "form data" encoded requests. In case of GET those are implied, but we deal with POST only in which case, this is just one of possible encodings. Not only that, it's also handled by more generic `request_body`: https://swagger.io/docs/specification/describing-request-body/

rslota: MongoosePush uses JSON "officially" and only that. HTTP's form data could be _allowed_ for APIv1, as it's flat anyway, but for APIv3 that's not even an option (ok, technically it is possible in a way that you described, but that's just hacky and it's one of the reasons why OpenAPI 3.0 is not compatible with Swagger 2.0).

michalwski: Ok, we have the path 'v1/notification/:device_id'
Do you mean that even though the device_id is part of path, it can be described in request_body ?

rslota: Oh wow, so sorry about that - the file preview "collapsed" a bit and I misread the implementation. No, you're completely right - we do have one query parameter and it should be handled the way it is. Again, sorry :)

lib/mongoose_push_web/schemas/api_v1_request_push.ex
```diff
@@ -0,0 +1,32 @@
...
+        enum: ["fcm", "apns"]
+      },
+      body: %Schema{type: :string, description: "Body of the notification", format: :string},
+      title: %Schema{type: :string, description: "Title of the notification", format: :string},
+      badge: %Schema{type: :integer, format: :int32},
+      click_action: %Schema{type: :string},
+      tag: %Schema{type: :string},
+      topic: %Schema{type: :string},
+      data: %Schema{type: :string},
+      mode: %Schema{type: :object, enum: [:prod, :dev]}
```
leszke: In Maru it was `Atom`, what is not available here, so I tried using `:object`.

lib/mongoose_push_web/controllers/api_v1_controller.ex
leszke: I do agree on keeping things in order and preventing any chaos here, especially regarding future development, but maybe should we stick for now with current file tree and develop it altogether with further PRs, as it's going to grow? Or you'd like to have it already set up :)

rslota: It's gonna be more work to refactor that later when we have more APIs defined. Also, I would argue that this change makes code more readable which affects the code even right now :)

lib/mongoose_push_web/schemas/api_v1_request_push.ex
leszke: I corrected the type to `string` to comply with JSON input standards, but still would like somebody to glance at it and check if that's correct.

lib/mongoose_push_web/schemas/request/send_notification/flat.ex
```diff
@@ -0,0 +1,32 @@
+defmodule MongoosePushWeb.Schemas.Request.SendNotification.Flat do
+  require OpenApiSpex
+  alias OpenApiSpex.Schema
+
+  OpenApiSpex.schema(%{
+    title: "APIv1.Request.Push",
+    description: "Push notification request schema",
+    type: :object,
+    properties: %{
```
michalwski, line 9: As far as I know `service`, `body` and `title` are required parameters. I think that `OpenApiSpex.Schema` allows to mark certain fields as required by providing a list of required fields like below:
```suggestion
    required: [:service, :body, :title],
    properties: %{
```

michalwski (commented):
Thanks for changing the code based on @rslota suggestions. Now, the structure and naming is more organized and easier to follow.
I had only one comment regarding required parameters. Other than that all looks good to me.

lib/mongoose_push_web/schemas/request/send_notification/flat.ex
michalwski: Oh, I didn't notice. It's added below `properties`. That's ok :)

michalwski (approved):

lib/mongoose_push_web/router.ex
```diff
@@ -5,9 +5,26 @@ defmodule MongoosePushWeb.Router do
...
     post("/dummy", DummyController, :handle)
   end
+
+  scope "/" do
+    pipe_through(:swagger_json)
+
+    get("/swagger.json", OpenApiSpex.Plug.RenderSpec, [])
+  end
+
+  scope "/v1", MongoosePushWeb do
```
rslota, line 25: This should be `MongoosePushWeb.APIv1`

leszke: Addressed by upcoming PR.

PR #146, opened 2020-04-01
MIM-865 minor fixes

leszke:
This spontaneous PR covers comments arisen during our remote-peer-review with @rslota after his coming back from holiday. Some naming alignments, more response codes for APIv1, added SwaggerUI for diagnostic purposes (to be hidden from production in near future!).


lib/mongoose_push_web/schemas/response/send_notification/generic_error.ex
```diff
@@ -0,0 +1,19 @@
...
+    type: :object,
+    properties: %{
+      details: %Schema{
+        type: :string,
+        description:
+          "Short description of the encountered error. For development/debugging purposes only, " <>
+            "not for any client usage, as it may change in time without prior notice."
+      }
+    },
+    example: %{}
```
rslota: We could add an example here (from tests?).

leszke: Done.

rslota (approved):

PR #149, opened 2020-04-21
MIM-867 Use OpenAPI to validate and cast Phoenix requests to structs

leszke:
This awaited PR addresses the story of making it possible to perform a proper (in terms of already developed APIs) push notification send request to phoenix backend HTTP endpoint (APIv1/APIv2/APIv3). Successful requests are replied with `200` status, while improper ones with `422`, including the brief description of what is not valid within the request.

In the home stretch I took the opportunity to remove unused code introduced in one of the former PRs, as well as did some minor alignments to already existing schemas definitions.

Known shortcomings of this PR:
- for APIv2/3 scenarios `OpenApiSpex` schemas validator does not return meaningful errors where expected - root cause has already been identified, workaround will follow soon;
- unit tests should be done in property-based testing manner - to be implemented in the upcoming sprint.


config/test.exs
```diff
@@ -1,6 +1,7 @@
 use Mix.Config
 
 config :mongoose_push, loglevel: :debug
+config :mongoose_push, mongoose_push: MongoosePushBehaviourMock
```
rslota: `mongoose_push` config option doesn't have any meaning. How about
`config :mongoose_push, backend_module: MongoosePushBehaviourMock`?

lib/mongoose_push.ex
```diff
@@ -13,6 +13,8 @@ defmodule MongoosePush do
   require Logger
   alias MongoosePush.Service
 
+  @behaviour MongoosePushBehaviour
```
rslota: This module does not conform to this behaviour yet. And if it does, then behaviour definition is incorrect (should expect correct structs that we pass from controller as input, while this module does not support it yet).

lib/mongoose_push_behaviour.ex
```diff
@@ -0,0 +1,5 @@
+defmodule MongoosePushBehaviour do
```
rslota: This shouldn't be a root level module, no module should be (I know we already have one, that was a mistake that I intend to fix soon).

lib/mongoose_push_behaviour.ex
```diff
@@ -0,0 +1,5 @@
...
+  @moduledoc false
+  @callback push(String.t(), MongoosePush.request()) ::
```
rslota: This spec is incorrect, the input is different from what controller actually passes. This module should define the input type according to want controller will / is passing.

lib/mongoose_push_web/controllers/api_v1_notification_controller.ex
```diff
@@ -5,6 +5,10 @@ defmodule MongoosePushWeb.APIv1.NotificationController do
 
   use MongoosePushWeb.Schemas
 
+  plug(OpenApiSpex.Plug.CastAndValidate)
+
+  def mongoose_push, do: Application.get_env(:mongoose_push, :mongoose_push)
```
rslota: That's a weird function name, how about `backend_module`?

rslota: Also, let's use `get_env!` or something similar, so that we crash when this setting is not set, instead crashing in weird `undef` reason.

rslota: This function also should be moved to `MongoosePush.Application` module, where all configuration is handled. This is especially useful, as we use this in 3 modules.

mix.exs
```diff
@@ -45,6 +45,7 @@ defmodule MongoosePush.Mixfile do
       {:lager, ">= 3.7.0", override: true},
       {:phoenix, "~> 1.4.16"},
       {:open_api_spex, "~> 3.6"},
+      {:mox, "~> 0.5.2"},
```
rslota: `only: [:test, :integration]`. Also, let's put in closer to other test-only deps.

test/mongoose_push_web/test_helper.exs
```diff
@@ -1,2 +1,6 @@
 {:ok, _} = Application.ensure_all_started(:mongoose_push)
 ExUnit.start(capture: true)
+
+defmodule Mocks do
```
rslota: Module here is redundant and actually may even cause issues. Let's never put modules in scripts (`.exs`).

rslota (changes requested):
Thanks for the PR! 
The general code seems to do whatever it suppose to do. The file structure / naming could use some work, also there are some other minor issues. 
I haven't go a chance to review tests - I'm gonna do that during next round of review.

config/test.exs
leszke: done

lib/mongoose_push.ex
leszke: done

lib/mongoose_push_behaviour.ex
leszke: done

lib/mongoose_push_behaviour.ex
leszke: done

lib/mongoose_push_web/controllers/api_v1_notification_controller.ex
leszke: done

mix.exs
leszke: done

test/mongoose_push_web/test_helper.exs
leszke: corrected, but the same was already done in `test/unit/test_helper.exs`

michalwski (approved):

PR #150, opened 2020-04-29
MIM-868 Handle APIv1 requests (phoenix controller)

leszke:
This story covers the topic of handling APIv1 compliant requests by our brand-new phoenix controller. Covering unit- and functional tests have also been added.

Known drawbacks:
- lack of property-based testing: it has been agreed this way of testing will follow in near future after assuring the main functionality of all the phoenix stories, as the design won't be trivial and will need some time to be spent on it;
- dialyzer error introduced by [this](https://github.com/esl/MongoosePush/blob/handle-apiv1-requests/lib/mongoose_push_web/controllers/api_v1/notification_controller.ex#L54) function call: after yesterday [investigation](https://esl-chat.slack.com/archives/CF3K60YTV/p1588082068098900) on slack channel I decided to escalate issue to PR level.


lib/mongoose_push_web/controllers/api_v1/notification_controller.ex
```diff
@@ -51,9 +51,13 @@ defmodule MongoosePushWeb.APIv1.NotificationController do
   end
 
   def send(conn = %{body_params: params}, %{device_id: device_id}) do
-    result = MongoosePush.Application.backend_module().push(device_id, params)
+    request = MongoosePushWeb.APIv1.RequestDecoder.decode(params)
+    result = MongoosePush.Application.backend_module().push(device_id, request)
     {status, payload} = MongoosePush.API.V1.ResponseEncoder.to_status(result)
 
+    # TODO remove the line below after the reimplementation of CastAndValidate plug
+    conn = update_in(conn.body_params, &Map.from_struct(&1))
```
kmakiela: I don't quite understand why we do that here

leszke: We are omitting dialyzer issues I mentioned in PR summary. More elaborative comment in `0b338fe9d2f3` description.

lib/mongoose_push.ex
```diff
@@ -13,6 +13,8 @@ defmodule MongoosePush do
   require Logger
   alias MongoosePush.Service
 
+  @behaviour MongoosePush.Notification
```
kmakiela, line 16: Can you elaborate on why we need this behaviour? As far as I understand it is only used for mocks in tests, is there any other reason?

leszke: It is the `Mox` library requirement to have behavior implemented by module in order to create its mock. You are correct we use it only for mocks.

lib/mongoose_push_web/schemas/request/send_notification/flat_notification.ex
```diff
@@ -45,4 +46,45 @@ defmodule MongoosePushWeb.Schemas.Request.SendNotification.Flat do
...
+
+      Enum.reduce(opt_alert_keys, push_request, fn x, acc ->
+        case Map.get(schema, x) do
+          nil -> acc
+          val -> Kernel.update_in(acc, [:alert, x], fn _ -> val end)
+        end
+      end)
+    end
+
+    defp maybe_convert_to_atom(:mode, val), do: String.to_atom(val)
```
michalwski: As far as I know, we are converting here any input from the user to an atom. I consider this unsafe. From what I remember we accept only two values (`prod` and `dev`) here so maybe we could just parse the input and return atom for these values we accept and rise and error or do sth esle appropriate for invalid values.

michalwski (changes requested):
Thanks for the changes. Elixir protocols indeed make sense here. The code looks good, I had only one strong objection `String.to_atom`, more detailed in the comment to the code.

Also, I know that property based tests are planned for future, I think it'd be good to add a test case for a case very the `mode` parameter in the request is set to sth else than `prod` or `dev` to check if correct error is returned.

lib/mongoose_push_web/schemas/request/send_notification/flat_notification.ex
rslota: `String.to_existing_atom/1`

michalwski: Even more, sth like:
```elixir
defp parse_mode("prod"), do :prod
defp parse_mode("dev), do :dev
```

leszke: thank you @michalwski for this catch

michalwski (approved):

PR #151, opened 2020-05-06
Telemetry metrics

kmakiela:
# Adding Telemetry Metrics support
* Changed telemetry events to support multi-dimensional metrics
* Added metrics' definitions
* Added `Prometheus` reporter
* Added `telemetry_metrics` and `telemetry_metrics_prometheus_core` dependencies
* Added tests for the metrics
* Changed config files to include `:service` -> module mapping
* Changed the loading of said module from hardcoded variable to
extracted from config
* Fixed `MongoosePush.Service.choose_pool/2` callback (it had arity of 1
and did not take tags as an argument)
* Added file defining mock with Mox (Needed because of compile-time
restraints)


kmakiela:
A note on what the Prometheus exposition formats look like:
https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -0,0 +1,124 @@
+defmodule MongoosePushTelemetryMetricsTest do
+  use ExUnit.Case, async: false
+  use Quixir
+  import Mock
```
rslota: Let's not use `Mock` in new tests - we'll be trying to get rid of this library. Use `Mox` instead.

lib/mongoose_push/application.ex
```diff
@@ -86,6 +88,39 @@ defmodule MongoosePush.Application do
     end)
   end
 
+  defp metrics do
+    [
+      # Summary is not yet supported in TelemetryMetricsPrometheus
```
rslota: Summary is not supported, indeed, but `Distribution` is. We should use it. Also, if counter ignores measurements, then  `mongoose_push.push.time` counter has very little sense :)

lib/mongoose_push.ex
```diff
@@ -140,9 +140,10 @@ defmodule MongoosePush do
 
   defp emit_telemetry_event(time, :ok, service, mode) do
     :telemetry.execute(
-      [:mongoose_push, service, :push, :success],
+      [:mongoose_push, :push],
```
rslota: Since we're changing the event name, can we change it to `[:mongoose_push, :notification, :send]`. I feel like this is more telling ;)

lib/mongoose_push/application.ex
```diff
@@ -86,6 +88,39 @@ defmodule MongoosePush.Application do
     end)
   end
 
+  defp metrics do
```
rslota: I feel like this should belong in some separate module dedicated for that. This will only get bigger.

lib/mongoose_push/application.ex
```diff
@@ -86,6 +88,39 @@ defmodule MongoosePush.Application do
...
+    [
+      # Summary is not yet supported in TelemetryMetricsPrometheus
+      Telemetry.Metrics.counter("mongoose_push.push.time",
+        tags: [:status, :service, :type, :reason],
+        tag_values: fn metadata ->
+          case metadata.status do
+            :success ->
+              Map.merge(
+                %{
+                  type: :success,
```
rslota: `error_category`?

lib/mongoose_push/application.ex
```diff
@@ -86,6 +88,39 @@ defmodule MongoosePush.Application do
...
+                  reason: :success
```
rslota: `error_reason`?

rslota: Also, I feel that empty value would be a better match here.

lib/mongoose_push/application.ex
rslota: In our code I guess the old name is fine without the `error_` prefix, but this will be external name, for someone that doesn't see the code. We need the context.

lib/mongoose_push/application.ex
```diff
@@ -86,6 +88,39 @@ defmodule MongoosePush.Application do
...
+                metadata
+              )
+          end
+        end
+      ),
+
+      # measurement is ignored in Counter metric
+      Telemetry.Metrics.counter("mongoose_push.supervisor.init.count", tags: [:service]),
+      Telemetry.Metrics.counter("mongoose_push.apns.state.init.count"),
+      Telemetry.Metrics.counter("mongoose_push.apns.state.terminate.count", tags: [:reason]),
```
rslota:
```suggestion
      Telemetry.Metrics.counter("mongoose_push.apns.state.terminate.count", tags: [:error_reason]),
```

rslota: If this is easier we can change the telemetry events to also use this naming, but as mentioned above:
```In our code I guess the old name is fine without the error_ prefix, but this will be external name, for someone that doesn't see the code. We need the context.```

**Review by rslota — CHANGES_REQUESTED** [review]

File: lib/mongoose_push/application.ex (hunk 5 above)
Comment (NelsonVides) [reply lib/mongoose_push/application.ex]: I really don't know what do you mean by `error_category` and where in the code you point out here this is supposed to refer to 😑

File: lib/mongoose_push/application.ex (hunk 6 above)
Comment (NelsonVides) [reply lib/mongoose_push/application.ex]: Likewise here, I don't get what is it about.

File: lib/mongoose_push/application.ex (hunk 7 above)
Comment (NelsonVides) [reply lib/mongoose_push/application.ex]: Likewise, can you please explain? I'm just more and more lost 😕

File: lib/mongoose_push/application.ex (hunk 5 above)
Comment (rslota) [reply lib/mongoose_push/application.ex]: Most of the functions in this applications return `{:error, {category(), reason()}}` error tuples. This means that error details are actually two atoms, not just one. 
The point I'm trying to make here, is that metadata called `type` attached to a metric that tracks pushed notification is at the very least confusing for reader. Please note, that people reading/operation on metrics have no idea how code looks like and what this `type` refers to. Here, I guess it's supposed to be "error type", so I guess in type above it would be `category()`. But even I knowing the codebase need to think for a moment what this field represents, I can imagine what someone without codebase or even Elixir knowladge would feel. 
This is often what happens, when we decide to "expose" some internal things to outside-world - we need to pay attention to naming, as when this data leaves the service, there is no context given, which renders the naming, very often, incorrcet.

File: test/unit/mongoose_push_telemetry_metrics_test.exs (hunk 1 above)
Comment (kmakiela) [reply test/unit/mongoose_push_telemetry_metrics_test.exs]: I see a problem here - Mox can mock behaviours only and the behaviour we want to mock is `MongoosePush.Service`, but we never directly call any functions by `Service.push`. We choose a module based on the `:service` type and then call the required function from a chosen module. Additionally, we have more than one implementation of said behaviour (`APNS` and `FCM`). Making different behaviours for different services kills the purpose of a behaviour in a first place tough. Considering all that, I don't quite see a clean way to mock what we need.

Comment (rslota) [reply test/unit/mongoose_push_telemetry_metrics_test.exs]: Then we simply need to move `:service` -> module mapping to configuration instead of having it hardcoded. That actually makes `:service` being more dynamic, which is also nice on it's own.

Comment (rslota) [reply test/unit/mongoose_push_telemetry_metrics_test.exs]: Why would we make more then one behavior? I don't think I get you point, can you elaborate?

Comment (rslota) [reply test/unit/mongoose_push_telemetry_metrics_test.exs]: Indeed, we have two implementations. Mock you be 3rd. 

We could change it so we have the following config file:

`prod.exs`:
```
config :mongoose_push, MongoosePush.Service:
  fcm: MongoosePush.Service.FCM,
  apns: MongoosePush.Service.APNS
```

`test.exs`:
```
config :mongoose_push, MongoosePush.Service:
  fcm: MongoosePush.Service.Mock,
  apns: MongoosePush.Service.Mock
```

Comment (kmakiela) [reply test/unit/mongoose_push_telemetry_metrics_test.exs]: The config file suggestion made it clear for me, thanks

Comment (kmakiela) [reply test/unit/mongoose_push_telemetry_metrics_test.exs]: Regarding that I decided to put
```
config :mongoose_push, MongoosePush.Service:
  fcm: MongoosePush.Service.FCM,
  apns: MongoosePush.Service.APNS
```
in `test.exs` as well, because it affects all tests and changing it should probably be a part of a story about replacing Mock with Mox

File: lib/mongoose_push/metrics/telemetry_metrics.ex [hunk 8]
```diff
@@ -0,0 +1,49 @@
...
+            :error ->
+              Map.merge(
+                %{
+                  error_category: Map.get(metadata, :type) || :generic,
+                  error_reason: metadata.reason
+                },
+                metadata
+              )
+          end
+        end
```
michalwski: Can these `tag_values` be prepared before the `Telemetry.Metrics.distribution` call?

michalwski (commented):
Thanks for the changes, looks like the metrics topic is not very straightforward.
I may be old-school or don't understand the multidimensional metrics very well. What we have now is one metric name for the notification sent event (`[:mongoose_push, :notification, :send]`) and some metadata inside indicating if the request succeeded or not. From (old-school) DevOps perspective I think it's more intuitive to have 2 metrics one for successful requests, the other for errors. When I have only one name, it requires more thinking when setting up the dashboard to distinguish from successes and errors.

I do understand that for example the error details could be set as the metadata, or the service name can also be part of the metadata. Please correct me where I'm wrong and make me less old-school :) 
CC @rslota

rslota:
> Thanks for the changes, looks like the metrics topic is not very straightforward.
> I may be old-school or don't understand the multidimensional metrics very well. What we have now is one metric name for the notification sent event (`[:mongoose_push, :notification, :send]`) and some metadata inside indicating if the request succeeded or not. From (old-school) DevOps perspective I think it's more intuitive to have 2 metrics one for successful requests, the other for errors. When I have only one name, it requires more thinking when setting up the dashboard to distinguish from successes and errors.
> 
> I do understand that for example the error details could be set as the metadata, or the service name can also be part of the metadata. Please correct me where I'm wrong and make me less old-school :)
> CC @rslota

That's a totally valid concern @michalwski :) In my opinion this is just a flip-of-coin judgment call whether status of operation should be part of metric name, or it's metadata. There is no good answer. However, in Prometheus specifically, tags (this metdata values) are part of a metric name, effectively. Measurements are and calculated for a given metrics in buckets defined by each unique set of tag values, so it almost doesn't matter where do you put the status of operation. The only difference is that, when querying, if you would like to get e.g. average time regardless of status, in case of status in tags, you would just put "*" as tag value, while if you had to separate metrics, you would need to write a query that sums two metrics and then calculates average - it would be longer and more complicated. So in my opinion, when deciding whether to put something to name or tag, you should ask yourself: "how likely is that I would need to concatenate data from those two metrics?". On the other hand, if you just need errors, then in both cases you would endup with very similar query, because if status is a tag, you would just add `{status=error}` to metric name.

michalwski:
Thanks for the explanation @rslota, this is clearer for me now. Let's have the name and tags as we have currently in this PR.

.dialyzer_ignore.exs
```diff
@@ -1,4 +1,6 @@
 [
+  ~r/lib\/mongoose_push\/metrics\/telemetry_metrics\.ex.*child_spec.*no local return/,
```
rslota, line 4: It would be nice to put a comment here why do we include those ignores. Maybe link to a bug report?

config/dev.exs
```diff
@@ -22,6 +22,10 @@ config :maru, MongoosePush.Router,
     otp_app: :mongoose_push
   ]
 
+config :mongoose_push, MongoosePush.Service,
```
rslota: We can put this in the "global" `config/config.exs` and just override it in test env. No need to duplicating it in both prod and dev.

lib/mongoose_push.ex
```diff
@@ -63,7 +63,7 @@ defmodule MongoosePush do
           :ok | {:error, Service.error()} | {:error, MongoosePush.error()}
   def push(device_id, %{:service => service} = request) do
     mode = Map.get(request, :mode, :prod)
-    module = MongoosePush.Application.services()[service]
+    module = Application.get_env(:mongoose_push, MongoosePush.Service) |> Keyword.get(service)
```
rslota: All configuration should be handled by `MongoosePush.Application` module - if we even would change method of configuration, it should be localized to only that module. Let's modify original `MongoosePush.Application.services()` instead.

rslota: I do understand that this will affect more then just `push` function on startup. But for this reason Mox has: https://hexdocs.pm/mox/Mox.html#stub_with/2 - to provide default implementation for functions that we do not plan on mocking.

rslota (changes requested):
The PR looks mostly good. There is however small issue with mock configuration - it happens only for specific actions (push) and not for others (pool startup). This may lead to confusion for other developer seeing some function being called on original modules, while other on mock modules (one of the things that Mox aims to "fix"). We should always inject mocked module as whole and make sure that implementation uses it in all cases, not only some. Also, this issue actually makes the configuration code "bleed" into application logic, which is also not preferable. Details as always in comments :)

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -0,0 +1,139 @@
...
+    MongoosePush.Service.Mock
+    |> expect(:push, repeat_no, fn _, _, _, _ -> push_result end)
+    |> stub_with(FCM)
+
+    ptest [
+            mode: choose(from: [value(:dev), value(:prod)]),
+            service: choose(from: [value(:fcm), value(:apns)])
+          ],
+          repeat_for: repeat_no do
+      assert push_result =
```
rslota: This is assignment, this assert will never fail, unless `push/2` returns `false` or `nil`. Maybe you meant:
```suggestion
      assert ^push_result =
```

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -0,0 +1,139 @@
...
+               push(
+                 "device_id",
+                 %{service: service, title: "", body: "", mode: mode}
+               )
+    end
+
+    Application.put_env(:mongoose_push, MongoosePush.Service,
```
rslota: We already have a function `TestHelper.reload_app()` to deal with resetting the app to default. I can see that it's  defined in one of the scripts - well it shouldn't be, so we can move it and use it.

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -0,0 +1,139 @@
...
+    metrics = TelemetryMetricsPrometheus.Core.scrape()
+    regex = ~r/mongoose_push_apns_state_get_default_topic_count (?<count>[\d]+)/
+    match = Regex.named_captures(regex, metrics)
+    count = get_count(match)
+
+    assert 10 == count
+  end
+
+  defp do_push(push_result, repeat_no) do
+    Application.put_env(:mongoose_push, MongoosePush.Service,
```
rslota: I feel like this kind of setting should always be part of setup block, explicitly.

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -0,0 +1,139 @@
...
+      fcm: MongoosePush.Service.Mock,
+      apns: MongoosePush.Service.Mock
+    )
+
+    MongoosePush.Service.Mock
+    |> expect(:push, repeat_no, fn _, _, _, _ -> push_result end)
+    |> stub_with(FCM)
+
+    ptest [
```
rslota: ptest i meant to be a test body, while this function is sometimes called in setup block. I'm not sure what kind of issues this may cause. 
Also, `ptest` is meant to be used as property based tester - in case the body fails, it's gonna try to find the failing value and print it out. Meanwhile, we're not using it as actual test body, so the output may be very surprising.
 I don't think this is critical, but it bothers me that we're using property-based testing framework to be effectively just a data generator. If we do that, then maybe we should consider just using the generators directly, not the `ptest` macro? I'm worried here, that using this framework for different purpose then it was built for may bite us.

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -0,0 +1,139 @@
...
+          Map.update(acc, [type, reason], count, fn val -> val + count end)
+        end)
+
+      assert 3 == length(Map.keys(errors))
+      assert true == Enum.all?(Map.values(errors), fn val -> val == 10 end)
+    end
+  end
+
+  test "APNS default topic extraction metrics" do
+    ptest [
```
rslota: Same here, we're miss-using `ptest` as generator. I honestly have no idea what implication it may have. Maybe let's just use `for` and https://hexdocs.pm/faker/api-reference.html#modules ?

rslota (changes requested):

rslota (approved):

test/unit/mongoose_push_telemetry_metrics_test.exs
```diff
@@ -10,6 +11,16 @@ defmodule MongoosePushTelemetryMetricsTest do
   setup do
     TestHelper.reload_app()
 
+    # TelemetryMetricsPrometheus starts asynchronously - we need to wait until metrics are
+    # registered, otherwise tests count run into race condition.
+    # Becasue of how TelemetryMetricsPrometheus startup is implemented right now, the following
+    # call alone will only return after the initialization (first one will be sucessful),
+    # but in case something would change in future we'll "wait" and retry until it returns
+    # at least one metric.
```
NelsonVides:
```suggestion
    # TelemetryMetricsPrometheus starts asynchronously - we need to wait until metrics are
    # registered, otherwise tests could run into a race condition.
    # Because of how TelemetryMetricsPrometheus startup is implemented right now, the following
    # call alone will only return after the initialization (first one will be successful),
    # but in case something would change in future we'll "wait" and retry until it returns
    # at least one metric.
```

PR #153, opened 2020-05-12
MIM-869 Handle APIv2 + APIv3 requests (phoenix controller)

leszke:
This pull request is the intuitive continuation of MIM-868 story, with the main difference that here I cover handling of the requests specified by APIv2 & APIv3.

In the home stretch it also turned out that I needed to - kind of - rollback to previous schemas design, reintroducing `MixedNotification` definition. Source rationale is that `OpenApiSpex` framework is not doing well when it finds incoming request matching more than one schema definition.

### DIFF
--- a/lib/mongoose_push/api/v2/response_encoder.ex
+++ b/lib/mongoose_push/api/v2/response_encoder.ex
@@ -5,7 +5,7 @@ defmodule MongoosePush.API.V2.ResponseEncoder do
   @behaviour MongoosePush.API
   alias MongoosePush.Service
 
-  @spec to_status(:ok | {:error, Service.error() | {:error, MongoosePush.error()}}) ::
+  @spec to_status(:ok | {:error, Service.error()} | {:error, MongoosePush.error()}) ::
           {non_neg_integer, %{details: atom | String.t()} | nil}
   def to_status(:ok), do: {200, nil}
 
--- a/lib/mongoose_push_web/schemas/request/send_notification/deep/mixed_notification.ex
+++ b/lib/mongoose_push_web/schemas/request/send_notification/deep/mixed_notification.ex
@@ -1,23 +1,30 @@
-defmodule MongoosePushWeb.Schemas.Request.SendNotification.Deep.SilentNotification do
+defmodule MongoosePushWeb.Schemas.Request.SendNotification.Deep.MixedNotification do
   require OpenApiSpex
   alias MongoosePushWeb.Protocols.RequestDecoder
   alias MongoosePushWeb.Schemas.Request.SendNotification.Deep
 
   OpenApiSpex.schema(%{
-    title: "Request.SendNotification.Deep.SilentNotification",
-    description: "In this request data field is mandatory.",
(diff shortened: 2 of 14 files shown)


lib/mongoose_push/api/v2/response_encoder.ex
```diff
@@ -5,7 +5,7 @@ defmodule MongoosePush.API.V2.ResponseEncoder do
   @behaviour MongoosePush.API
   alias MongoosePush.Service
 
-  @spec to_status(:ok | {:error, Service.error() | {:error, MongoosePush.error()}}) ::
+  @spec to_status(:ok | {:error, Service.error()} | {:error, MongoosePush.error()}) ::
```
rslota, line 8: Nice catch!

lib/mongoose_push_web/schemas/request/send_notification/deep/mixed_notification.ex
```diff
@@ -0,0 +1,36 @@
...
+  })
+
+  defimpl RequestDecoder,
+    for: MongoosePushWeb.Schemas.Request.SendNotification.Deep.MixedNotification do
+    alias MongoosePushWeb.Protocols.RequestDecoderHelper
+
+    @spec decode(%Deep.MixedNotification{}) :: MongoosePush.request()
+    def decode(schema) do
+      %{
+        service: String.to_atom(schema.service),
```
rslota: Don't we have a helper that does this better? Even if not, then let's use `to_existing_atom` at least.

lib/mongoose_push_web/schemas/request/send_notification/deep/silent_notification.ex
```diff
@@ -1,14 +1,29 @@
...
   })
+
+  defimpl RequestDecoder,
+    for: MongoosePushWeb.Schemas.Request.SendNotification.Deep.SilentNotification do
+    alias MongoosePushWeb.Protocols.RequestDecoderHelper
+
+    @spec decode(%Deep.SilentNotification{}) :: MongoosePush.request()
+    def decode(schema) do
+      %{
+        service: String.to_atom(schema.service),
```
rslota: Don't we have a helper that does this better? Even if not, then let's use to_existing_atom at least.

rslota (commented):
The PR looks great! There is one small issue with `String.to_atom` use, but other then that, ready to be merged :)

lib/mongoose_push_web/schemas/request/send_notification/deep/mixed_notification.ex
leszke: I implemented one, following the way @michalwski already suggested [here](https://github.com/esl/MongoosePush/pull/150/#discussion_r421375560).

lib/mongoose_push_web/schemas/request/send_notification/deep/silent_notification.ex
leszke: same as above

rslota (approved):

PR #155, opened 2020-05-13
Integration test for prometheus metrics

NelsonVides:
This mess works on my computer.

### DIFF
--- a/test/integration/prometheus_endpoint_test.exs
+++ b/test/integration/prometheus_endpoint_test.exs
@@ -1,0 +1,42 @@
+defmodule MongoosePush.API.PrometheusEndpointTest do
+  use ExUnit.Case, async: false
+  use AssertEventually, timeout: 5000, interval: 10
+
+  alias MongoosePush.Support.API, as: Tools
+
+  @url "/v1/notification/f534534543"
+
+  setup do
+    Tools.reset(:fcm)
+    :ok
+  end
+
+  @tag :skip
+  test "Metrics can be correctly received from Prometheus" do
+    # Push some notification to trigger metrics
+    notification = %{
+      :service => :fcm,
+      :body => "A message from someone",
+      :title => "Notification title"
+    }
+
+    assert {200, _} = Tools.post(@url, notification)
+
+    # Fetch the metrics
+    {status, headers, metrics} = Tools.get("/metrics")
+
+    # 1. Status is 200
+    assert 200 == status
+
+    # 2. content type is text/plain
+    {"content-type", resp} = List.keyfind(headers, "content-type", 0)
+    assert String.contains?(resp, "text/plain")
+
+    # 3. regex on the payload to make sure this is prometheus output
+    fcm_regex =
+      ~r/mongoose_push_notification_send_time_count{error_category=\"\",error_reason=\"\",service=\"fcm\",status=\"success\"} (?<count>[\d]+)/
+
+    fcm_match = Regex.named_captures(fcm_regex, metrics)
+    assert 0 != fcm_match
+  end
+end
--- a/test/support/api.ex
+++ b/test/support/api.ex
@@ -28,6 +28,13 @@ defmodule MongoosePush.Support.API do
     }
   end
 
+  def get(path) do
+    %Response{status_code: status, headers: headers, body: body} =
+      HTTPoison.get!("https://localhost:8443" <> path)
+
+    {status, headers, body}
+  end
+
   def post(path, json) do
     %Response{status_code: status_code, body: body} =
       HTTPoison.post!(


test/integration/connection_lost_test.exs
```diff
@@ -27,4 +27,37 @@ defmodule MongoosePush.API.ConnectionTest do
       Process.sleep(1000)
     end)
   end
+
+  test "Metrics can be correctly received from Prometheus" do
```
rslota: We should make new test suite (file) for this test. It hardly qualifies as "connectio_lost_test".

test/integration/connection_lost_test.exs
```diff
@@ -27,4 +27,37 @@ defmodule MongoosePush.API.ConnectionTest do
...
+    # Push some notification to trigger metrics
+    %{status_code: status_code, body: body} =
+      HTTPoison.post!(
```
rslota: I think we already have helpers that do such requests in `MongoosePush.Support.API` module.

test/integration/connection_lost_test.exs
```diff
@@ -27,4 +27,37 @@ defmodule MongoosePush.API.ConnectionTest do
...
+        "https://localhost:8446/v1/notification/f53453455",
+        "{\"badge\":7,\"body\":\"A message from someone\",\"click_action\":\".SomeApp.Handler.action\",\"data\":{\"custom\":\"data fields\",\"nested\":{\"fields\":\"allowed\"},\"some_id\":345645332},\"mode\":\"prod\",\"service\":\"apns\",\"tag\":\"info\",\"title\":\"Notification title\",\"topic\":\"com.someapp\"}",
```
rslota: The payload should be a map that is encoded to JSON with a library, for better readability.

test/integration/connection_lost_test.exs
```diff
@@ -27,4 +27,37 @@ defmodule MongoosePush.API.ConnectionTest do
...
+        [{"Content-Type", "application/json"},
+         {"accept", "application/json"}],
+        hackney: [:insecure]
+      )
+    assert {200, _} = {status_code, body}
+
+    # Fetch the metrics
+    {:ok, %HTTPoison.Response{status_code: status, body: metrics, headers: headers}} =
+      HTTPoison.get("https://localhost:8446/metrics")
```
rslota: Same here, we should use / modify `MongoosePush.Support.API` so that we have only one place where host/port is setup.

test/integration/connection_lost_test.exs
```diff
@@ -27,4 +27,37 @@ defmodule MongoosePush.API.ConnectionTest do
...
+
+    # 1. Status is 200
+    assert 200 == status
+
+    # 2. content type is text/plain
+    IO.inspect headers
```
rslota: I guess it's a debug thingy to remove :)

test/integration/connection_lost_test.exs
rslota: Also we need to disable this test (`@tag :skip`).

rslota (changes requested):

test/integration/prometheus_endpoint.exs
```diff
@@ -0,0 +1,42 @@
...
+  alias MongoosePush.Support.API, as: Tools
+
+  @url "/v1/notification/f534534543"
+
+  setup do
+    Tools.reset(:fcm)
+    :ok
+  end
+
+  # @tag :skip
```
rslota: This test is not being picked up by ExUnit because it doesn't have the `_test.exs` suffix. We should rename that file and enable the skip that should be required after the rename.

test/support/api.ex
```diff
@@ -28,6 +28,13 @@ defmodule MongoosePush.Support.API do
     }
   end
 
+  def get(path) do
+    {:ok, %Response{status_code: status, headers: headers, body: body}} =
```
rslota: The `post/2` function uses `HTTPoison.post!`, while here we use pattern match with `HTTPoison.get` (no bang!). I don't mind either approach, but it would be nice to keep those functions behavior the same.

rslota (changes requested):
The PR looks great, other then wrong filename for test suite. After we fix that we're good to go!

test/support/api.ex
rslota: This is gonna crash now as !bang version does not return ok-tuple ;)

test/integration/prometheus_endpoint_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  alias MongoosePush.Support.API, as: Tools
+
+  @url "/v1/notification/f534534543"
+
+  setup do
+    Tools.reset(:fcm)
+    :ok
+  end
+
+  # @tag :skip
```
rslota: As expected, integration tests failed. We need to uncomment this.

rslota (changes requested):

rslota (approved):

test/integration/prometheus_endpoint_test.exs
```diff
@@ -0,0 +1,42 @@
...
+
+  @url "/v1/notification/f534534543"
+
+  setup do
+    Tools.reset(:fcm)
+    :ok
+  end
+
+  @tag :skip
+  test "Metrics can be correctly received from Prometheus" do
```
michalwski, line 15: @rslota @NelsonVides, can this test case be enabled now when Phoenix is finally used instead of Maru?

rslota: @michalwski I've already made that change here: https://github.com/esl/MongoosePush/pull/159/files#diff-74898c36c774be8df73c05fbcd46d647L14

PR #157, opened 2020-05-21
Dockerfile refactor

rslota:
* Use single `Dockerfile` for build and release for simplicity 
* Use the same `Dockefile` on CI
* Updated docs


.circleci/config.yml
```diff
@@ -184,32 +206,40 @@ jobs:
...
       - run:
           name: Login to DockerHub
           command: docker login -u "${DOCKERHUB_USER}" -p "${DOCKERHUB_PASS}"
       - run:
           name: Build docker image
-          command: docker build -f Dockerfile.release --build-arg MIX_ENV=prod -t $DOCKERHUB_REPOSITORY/$DOCKER_IMAGE:$DOCKER_TAG .
+          command: docker build -f Dockerfile --build-arg MIX_ENV=prod -t mpush:latest .
+      - run:
+          name: Push docker image but do not tag it as the latest
+          command: docker tag mpush:latest $DOCKERHUB_REPOSITORY/$DOCKER_IMAGE:$DOCKER_TAG
```
michalwski, line 218: Would it be possible and make sense to push the docker image only after integration tests pass?

Dockerfile
```diff
@@ -1,5 +1,31 @@
+FROM rslota/beam-builder:erlang-22.0_elixir-1.9 AS builder
```
michalwski, line 1: Can we move the `beam-builder` docker image under `mongooseim` or `esl` org on dockerhub?

michalwski (commented):
Thanks for the changes. It will be easier to maintain the docker files and have working test env on local machines with these changes. 
Also, I hade 2 questions on the code.

.circleci/config.yml
rslota: Actually the whole point here was to make it possible for developer to run this docker image locally and see why tests fail, without need for local image building.

Dockerfile
rslota: Yeah, we can.

.circleci/config.yml
rslota: One thing I was thinking of here is to maybe create a separate dockerhub repo for those "development" images.

michalwski: All right, that explains. In the end even if tests fails, usually the issue is fixed in the PR so the broken image will be overwritten.

Dockerfile
rslota: But to avoid miscommunication - this is big enough in terms of waiting time, that definitely not as part of this PR, as it blocks quite a few things.

leszke (approved):

PR #160, opened 2020-05-22
Exometer extermination

leszke:
In this PR we get rid of `exometer` dependency, which will no longer be used in the project. It was agreed with @rslota that documentation update will follow in one of the subsequent PRs, being the part of the same story.


README.md
```diff
@@ -332,36 +332,6 @@ If you specify both **alert** and **data**, target device will receive both noti
...
-
-or if you modify an existing release (`sys.config`):
-```erlang
-{exometer_core,[{report, [{reporters, [{exometer_report_graphite, GRAPHITE_OPTIONS}]}]}]},
-{elixometer,
-    [{reporter, exometer_report_graphite},
-     {env, prod},
-     {metric_prefix, <<"mongoose_push">>}]},
-```
-
```
rslota, line 360: Sections below are also not really applicable right now and also should be removed (so they can be updated). "I use MongoosePush docker, where do I find `sys.config`" section is now not really need, maybe apart from "very advanced customizations", but I think I'm gonna cover that when introducing new config file. The section "Available metrics" is outdated, and needs to be rewritten, a metric names have changed.

rslota (changes requested):

rslota (approved):

PR #163, opened 2020-05-25
Reenable bind_addr setting

rslota:


leszke (approved):

config/prod.exs
```diff
@@ -6,7 +6,13 @@ config :goth, endpoint: {:system, :string, "FCM_AUTH_ENDPOINT", "https://www.goo
...
 config :mongoose_push, MongoosePushWeb.Endpoint,
   https: [
-    ip: {0, 0, 0, 0},
+    ip: {
+      :system,
+      # Custom type parser (Phoenix needs erlang-inet-style IP address)
+      {MongoosePush.Config.ConfexProvider, :parse_bind_addr, []},
+      "PUSH_HTTPS_BIND_ADDR",
+      {127, 0, 0, 1}
+    },
```
michalwski, line 15: The way I understand the code in this format is that if the env var `PUSH_HTTPS_BIND_ADDR` is not provided for the `prod` release, Phoenix endpoint will be bound only to the `localhost` interface. Do I understand it correctly, is this the desired default behaviour?

michalwski (commented):
Thanks for reenabling the `bind_addr`. I just had one question, see in the code.

config/prod.exs
rslota: Yes, you understand correctly. And yes, this IMHO should be default behavior (and it was since beginning of mpush) - this is because, this simply is the "safer" choice, that should only be changed if the deploy environment need it. For example, docker containers pretty much **require** this to be set to `0.0.0.0`, as you never run more the one service in a single container. That's why in our dockerfile we have: https://github.com/esl/MongoosePush/blob/e9b9f7d08c8b14d0f6d4193932fffb96e68d138c/Dockerfile#L52

If you think this is inconvenient, we could change the default, but as separate PR. I just reverted the previous state here.

michalwski: That's fair enough, thanks for the explanation!

PR #164, opened 2020-05-26
Add support for FCM mutable-content

rslota:
Not tested!


NelsonVides:
@kamilwaz can you untag me as a reviewer here? I've just noticed this and can't untag me myself as I'm not a member of the repo 😄

PR #166, opened 2020-05-26
Configurable swagger

leszke:
The purpose of this PR is to make both swagger.json and swaggerUI configurable (on/off) from the application's config level.


lib/mongoose_push_web/router.ex
```diff
@@ -8,14 +8,14 @@ defmodule MongoosePushWeb.Router do
 
   pipeline :swagger_json do
     plug(:accepts, ["json"])
-    plug(OpenApiSpex.Plug.PutApiSpec, module: MongoosePushWeb.ApiSpec)
+    plug(MongoosePushWeb.SwaggerPlug, module: MongoosePushWeb.ApiSpec)
```
rslota:
```suggestion
    plug(MongoosePushWeb.Plug.MaybePutApiSpec, module: MongoosePushWeb.ApiSpec)
```

This is not technically swagger, this is OpenAPI. And the original naming is actually quite nice, as this plug "puts" the spec into the endpoint. There is also plug that validates agains it, so `SwaggerPlug` is simply too generic in my opinion.

lib/mongoose_push_web/swagger_plug.ex
```diff
@@ -0,0 +1,23 @@
+defmodule MongoosePushWeb.SwaggerPlug do
+  use Plug.Builder
+
+  @behaviour Plug
+
+  @impl Plug
+  def init(opts) do
+    if(Application.get_env(:mongoose_push, :enable_swagger) == true) do
```
rslota:
```suggestion
    if Application.get_env(:mongoose_push, :enable_swagger) do
```

rslota: Also, let's change `:enable_swagger` option to something more specific, like:

```
 config :mongoose_push, openapi: 
  expose_spec: true
```

Then will be also able to add another options, like:

 ```
 config :mongoose_push, openapi: 
  expose_spec: true,
  expose_ui: false
```

File: lib/mongoose_push_web/swagger_plug.ex [hunk 3]
```diff
@@ -0,0 +1,23 @@
+defmodule MongoosePushWeb.SwaggerPlug do
+  use Plug.Builder
```
rslota: Do we actually use that?

lib/mongoose_push_web/swagger_plug.ex
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  @impl Plug
+  def call(conn, nil) do
+    conn
+    |> put_status(404)
+    |> halt()
+  end
+
+  def call(conn, opts) do
```
rslota: `@impl Plug`

lib/mongoose_push_web/swagger_plug.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  @impl Plug
+  def init(opts) do
+    if(Application.get_env(:mongoose_push, :enable_swagger) == true) do
+      OpenApiSpex.Plug.PutApiSpec.init(opts)
+    end
+  end
+
+  @impl Plug
+  def call(conn, nil) do
```
rslota: What if `OpenApiSpex.Plug.PutApiSpec.init/1` returns nil? We can't really know for sure, as this is not really documented, and thus may change. We should return from our `init/1` something very specific, maybe very specific atom, and match on that instead.

rslota (changes requested):

lib/mongoose_push_web/swagger_plug.ex
leszke: changed to `import Plug.Conn`

lib/mongoose_push_web/swagger_plug.ex
leszke: You're right, but when user defines flag improperly, eg. `:false`, the `if` expression will be still evaluated as true. Isn't that something we should secure?

lib/mongoose_push_web/swagger_plug.ex
leszke: Currently `init` returns `nil` when the flag is set to `false`, so we're not creating all the stuff when we don't expect it to happen.

lib/mongoose_push_web/swagger_plug.ex
rslota: Actually, `:false` is exactly the same as `false` as the latter one is just an alias to the first one. But I get your point, I still think this is too defensive and looks weird, but if you feel strongly about it, let's leave it ;)

lib/mongoose_push_web/swagger_plug.ex
rslota: Yeah, but what if `OpenApiSpex.Plug.PutApiSpec.init/1` starts returning `nil`?

lib/mongoose_push_web/swagger_plug.ex
leszke: Well yeah you're absolutely right, I mess something up there while I was writing the comment.

lib/mongoose_push_web/swagger_plug.ex
leszke: Then currently this `nil` is propagated like in the previous case. Will change it.

lib/mongoose_push_web/plugs/maybe_put_swagger_ui.ex
```diff
@@ -0,0 +1,30 @@
...
+    opts =
+      if enable do
+        OpenApiSpex.Plug.SwaggerUI.init(opts)
+      end
+
+    %{enabled: enable, opts: opts}
+  end
+
+  @impl Plug
+  def call(conn, %{enabled: enable, opts: opts}) when enable == false or opts == nil do
```
rslota: Why do we check `opts` here? Is that documented that `nil` is a failure scenario? 
Actually, even if it is, the plug that generates this value should be responsible for handling it.

lib/mongoose_push_web/plugs/maybe_render_spec.ex
```diff
@@ -0,0 +1,28 @@
...
+    opts =
+      if enable do
+        OpenApiSpex.Plug.RenderSpec.init(opts)
+      end
+
+    %{enabled: enable, opts: opts}
+  end
+
+  @impl Plug
+  def call(conn, %{enabled: enable, opts: opts}) when enable == false or opts == nil do
```
rslota: Same as above regarding `opts`.

lib/mongoose_push_web/plugs/maybe_put_swagger_ui.ex
rslota: Also, there is not need for guard, let's just match on `enabled: false`.

lib/mongoose_push_web/plugs/maybe_render_spec.ex
```diff
@@ -0,0 +1,28 @@
...
+    conn
+    |> send_resp(404, "swagger.json access disabled")
```
rslota: This plug returns JSON by default. If we want to return a text, we should make sure the content type and format of the response is correct.

config/dev.exs
```diff
@@ -1,7 +1,13 @@
 use Mix.Config
 
 config :mongoose_push, loglevel: :debug
-config :mongoose_push, enable_swagger: true
+
+config :mongoose_push,
+  openapi: [
+    expose_spec: true,
+    expose_ui: false
```
rslota: In dev mode we should expose everything.
We also miss prod config.

lib/mongoose_push_web/plugs/maybe_put_swagger_ui.ex
```diff
@@ -0,0 +1,30 @@
+defmodule MongoosePushWeb.Plug.MaybePutSwaggerUI do
+  import Plug.Conn
+  @behaviour Plug
+
+  @impl Plug
+  def init(opts) do
+    enable =
+      Application.get_env(:mongoose_push, :openapi)[:expose_spec] &&
```
rslota: Maybe it would be cleaner to fetch `Application.get_env(:mongoose_push, :openapi)` once, assign it and just access sub-fields twice?

rslota (changes requested):

lib/mongoose_push_web/plugs/maybe_put_swagger_ui.ex
leszke: I've just pushed the update you're asking for :D

lib/mongoose_push_web/plugs/maybe_render_spec.ex
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+    %{enabled: enable, opts: opts}
+  end
+
+  @impl Plug
+  def call(conn, %{enabled: false}) do
+    conn
+    |> put_resp_content_type("application/json")
+    |> send_resp(404, "{\"error\":\"swagger.json disabled\"}")
```
rslota: Let's encode JSON with proper JSON encoder, not by hand ;)

test/docker/docker-compose.mpush.yml
```diff
@@ -25,6 +25,8 @@ services:
       - PUSH_FCM_ENDPOINT=fcm-mock
       - TLS_SERVER_CERT_VALIDATION=false
       - FCM_AUTH_ENDPOINT=http://fcm-mock:4001
+      - PUSH_OPENAPI_EXPOSE_SPEC=false
```
rslota: This file is used for tests on CI and local ones. Let's enable those two by default here.

rslota (changes requested):
The PR looks good and is almost ready to be merged! Only 2 very minor things ;)

NelsonVides (approved):

rslota (approved):

PR #168, opened 2020-06-02
Implement openapispex plug

leszke:
This PR introduces extraordinary way of handling incoming requests by `MongoosePush` altogether with `OpenApiSpex`.

The thing is, broadly speaking, that when `OpenApiSpex` validators match incoming request to the schema defined as `oneOf: [%Reference{}]`, and the request matches more than one alternative schema, the framework likes to ultimately throw misleading error `Failed to cast value to one of: [] (no schemas provided)`, instead of something more meaningful. Moreover, this behavior is approved by `OpenApiSpex` creators, and is reproducable by [this test](https://github.com/open-api-spex/open_api_spex/blob/master/test/cast/one_of_test.exs#L62). Here we avoid it by manual, on-the-fly modification of the pattern the request is matched against, to consist of purely one schema without `oneOf` keyword, so the framework can serve us as expected.

Figuratively speaking, before:
```
$ curl --header "Content-Type: application/json" --data '{"alert" : {"title" : "title", "body" :""}}' https://localhost:8443/v2/notification/test -k
{"errors":[{"message":"Failed to cast value to one of: [] (no schemas provided)","source":{"pointer":"/"},"title":"Invalid value"}]}%
$ curl --header "Content-Type: application/json" --data '{"service" : "apns"}' https://localhost:8443/v3/notification/test -k
{"errors":[{"message":"Failed to cast value to one of: [] (no schemas provided)","source":{"pointer":"/"},"title":"Invalid value"}]}%
```
After:
```
$ curl --header "Content-Type: application/json" --data '{"alert" : {"title" : "title", "body" :""}}' https://localhost:8443/v2/notification/test -k
{"errors":[{"message":"Missing field: service","source":{"pointer":"/service"},"title":"Invalid value"}]}%
$ curl --header "Content-Type: application/json" --data '{"service" : "apns"}' https://localhost:8443/v3/notification/test -k
{"errors":[{"message":"Missing field: alert","source":{"pointer":"/alert"},"title":"Invalid value"}]}%
```
I have also removed some naughty dialyzer trick from `APIv1.NotificationController`, as it was no longer needed.


lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+  def init(opts) do
+    OpenApiSpex.Plug.CastAndValidate.init(opts)
+  end
+
+  @impl Plug
+  def call(conn = %{params: %{"service" => _, "alert" => _}}, opts = %{operation_id: _}) do
+    OpenApiSpex.Plug.CastAndValidate.call(conn, opts)
+  end
+
+  @impl Plug
```
aleklisi: Is `@impl Plug` needed for each head or just for the 1st function head?

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+          Access.key(:content),
+          "application/json",
+          Access.key(:schema)
+        ],
+        fn %OpenApiSpex.Schema{oneOf: _listOfSchemas} ->
+          %OpenApiSpex.Reference{
+            "$ref": "#/components/schemas/Request.SendNotification.Deep.SilentNotification"
+          }
+        end
+      )
```
michalwski: As far as I can tell this block of code differs from the other `Kernel.update_in` by only the reference to schema. Do you think these two calls to `Kernel.update_in` can be refactored in a way that the reference is passed as an argument to a common function? Would such refactoring make sense?

michalwski (commented):
Thanks for implementing the custom Plug on top of `OpenApiSpex.Plug.CastAndValidate`. In our case, we can now have better error codes while still accepting different request schemas. I had only one comment on the code.

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+    update_schema_and_do_call(conn, opts)
+  end
+
+  @impl Plug
+  def call(conn = %{params: %{"alert" => _}}, opts = %{operation_id: _}) do
+    update_schema_and_do_call(conn, opts)
+  end
+
+  @impl Plug
+  def call(conn = %{params: %{"data" => _}}, opts = %{operation_id: _}) do
```
aleklisi: Why do you need to match on `operation_id`?  You do not use it anyway.

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+            {private_data.spec, operation_lookup}
+          )
+
+          operation
+
+        operation ->
+          operation
+      end
+
+    if operation.operationId do
```
aleklisi: Is it possible that `operationId` does not exist in the operation map?
If it does not always exist it will throw an error:
```elixir
iex(1)> m = %{a: 1}
%{a: 1}
iex(2)> m.a
1
iex(3)> m.b
** (KeyError) key :b not found in: %{a: 1}
iex(3)> m[:a]
1
iex(4)> m[:b]
nil
```
since you raze the error yourself, I assume you do not want access operator to do it for. Am I right?
```suggestion
    if operation[:operationId] do
```

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
+defmodule MongoosePushWeb.Plug.CastAndValidate do
+  @behaviour Plug
```
aleklisi, line 7: Does it make sense to add `moduledoc` to this plug?
https://hexdocs.pm/elixir/writing-documentation.html

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+
+  @impl Plug
+  def init(opts) do
+    OpenApiSpex.Plug.CastAndValidate.init(opts)
```
aleklisi, line 11: How about aliasing `OpenApiSpex.Plug.CastAndValidate` to `OpenApiSpexValidator` or something similar?

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+  # so the framework can serve us as expected.
+  defp update_schema_and_do_call(
+         conn = %{params: %{"data" => _}},
+         opts = %{operation_id: operation_id}
+       ) do
+    conn =
+      Kernel.update_in(
+        conn,
+        [
+          Access.key(:private),
```
aleklisi: How about extracting the list to a separate function parameterized with `operation_id`?
```elixir
defp schema_access_path(operation_id) do
    [
          Access.key(:private),
          :open_api_spex,
          :operation_lookup,
          operation_id,
          Access.key(:requestBody),
          Access.key(:content),
          "application/json",
          Access.key(:schema)
        ]
 end
```
Since it is used twice.

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+      raise "operationId was not found in action API spec"
+    end
+  end
+
+  @impl Plug
+  def call(conn, opts) do
+    OpenApiSpex.Plug.CastAndValidate.call(conn, opts)
+  end
+
+  # When OpenApiSpex matches incoming request to schema defined as `oneOf: [%Reference{}]`,
```
aleklisi: This can be a `@doc` or `@moduledoc` whichever you find more appropriate.

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+          }
+        end
+      )
+
+    OpenApiSpex.Plug.CastAndValidate.call(conn, opts)
+  end
+
+  defp update_schema_and_do_call(conn, opts = %{operation_id: operation_id}) do
+    conn =
+      Kernel.update_in(
```
aleklisi:
```suggestion
      update_in(
```
Kernel functions are always available.

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
```diff
@@ -27,7 +27,7 @@ defmodule MongoosePushWeb.APIv3NotificationControllerTest do
         Jason.encode!(Map.drop(ControllersHelper.alert_request(), ["service"]))
       )
 
-    assert json_response(conn, 422) == ControllersHelper.no_schemas_provided_response()
+    assert json_response(conn, 422) == ControllersHelper.missing_field_response("service")
```
aleklisi, line 30: Those tests look very generic. Maybe there is a way of generating them.
```elixir
@test_cases [
    {["service"], "service"},
    {["alert"], "alert"},
...
]
for {dropped, missing} <- @test_cases do
    test "Request.SendNotification.Deep.AlertNotification schema without required #{missing} field", %{ conn: conn } do
        conn =post(conn, "/v3/notification/123456", Jason.encode!
        (Map.drop(ControllersHelper.alert_request(), ["alert"])))

        assert json_response(conn, 422) == ControllersHelper.missing_field_response("alert")
    end
end
```
formatting might be fixed

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
```diff
@@ -40,7 +40,7 @@ defmodule MongoosePushWeb.APIv3NotificationControllerTest do
         Jason.encode!(Map.drop(ControllersHelper.alert_request(), ["alert"]))
```
aleklisi, line 40: How about build a body in a separate variable and then use it?
```elixir
body = 
    ControllersHelper.alert_request()
    |> Map.drop(["alert"])
    |> Jason.encode!()
```

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
```diff
@@ -105,7 +105,7 @@ defmodule MongoosePushWeb.APIv3NotificationControllerTest do
         Jason.encode!(Map.drop(ControllersHelper.silent_request(), ["data"]))
```
aleklisi, line 105: So you drop `data`, but get that `alert is missing`?

aleklisi (commented):
Nice PR.
You use both `operation_id` and `operationId` in your code can you please make it consistent across the code. I assume `operation_id` is more suitable.
Please add test what happens if you drop both `alert` and `service` and another one where you drop all 3 (with `data` too).
Why there are no API v1 tests for this functionality? Will you please add them?
Regarding tests, you can autogenerate them with
```elixir
@apis [ {"v1", "/v1/notification/654321"}, ...]
@test_cases [ ... ]
for {api_name, endpoint} <- @apis do
    for ... <- @test_cases do
        test ...
    end
end
```
the ... parts are included in one of the comments to the PR.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: Yes it definitely would. Thanks for catching.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: I think it should be added and helps to keep the module consistent.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: Because the `operation_id` option is introduced only after the first call of `call/2` function (the clause starting from `call(conn = %{private: %{}...})`, where it is read from `conn.private.open_api_spex` field content, so the function calls flow remains correct. The content of this file follows the original plug implementation.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: For my best knowledge it should always exist at this point of the request flow. In case it couldn't, I don't see any need for such a defensive programming anyway. Moreover, this part of code is also copied from the original plug. Please refer to `OpenApiSpex.Plug.CastAndValidate` file content.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: It's a wrapper for already documented `OpenApiSpex.Plug.CastAndValidate` module, so I would find more than 1 line doc reduntant here.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: I personally don't think it will improve the readability, as currently I strictly differentiate my wrapping `MongoosePushWeb.Plug.CastAndValidate` from the original plug `OpenApiSpex.Plug.CastAndValidate`.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: That's almost exactly what @michalwski comment above is about.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: From the hexdoc you linked above:

> Elixir treats documentation and code comments as different concepts. Documentation is an explicit contract between you and users of your Application Programming Interface (API), be them third-party developers, co-workers, or your future self. Modules and functions must always be documented if they are part of your API.

> **Code comments** are aimed at developers reading the code. They are useful for marking improvements, **leaving notes (for example, why you had to resort to a workaround due to a bug in a library)**, and so forth. They are tied to the source code: you can completely rewrite a function and remove all existing code comments, and it will continue to behave the same, with no change to either its behaviour or its documentation.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: They are, but here you are always sure which function you are calling. The `Kernel` one, not some other private helper.

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
leszke: It is indeed a good idea, and since the property-based testing story is the next one in my backlog, I didn't spend so much time on it since it will be rewritten anyway.

leszke: It is indeed a good idea, and since the property-based testing story is the next one in my backlog, I didn't spend so much time on it cause it will be rewritten anyway.

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
leszke: The matter of preference. To be considered when doing property-based stuff :)

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
leszke: Paradoxically yes, because after this drop, the incoming request - from all the mandatory fields - has only `service` field, and in the flow we developed yesterday I match it against `AlertNotification` to simplify the things. The request incoming is just a freshly parsed map, and the whole struct thing is only a conjecture.
This comment is a good catch, but like I mentioned somewhere above - I don't want to waste so much time on tests that are next in line to be rewritten after this story is closed.

leszke:
And since I cannot reply to the general comment above, here goes my replies:
> You use both operation_id and operationId in your code can you please make it consistent across the code. I assume operation_id is more suitable.

This module takes a lot from original `CastAndValidate` plug source (I couldn't find any better way to impact its behavior only partially without copying things). I didn't want to change it simply because you can see there are no changes in how its functions work, I only wrap some additional stuff around them to enforce the behavior I expect.

> Please add test what happens if you drop both alert and service and another one where you drop all 3 (with data too).

The first scenario is already somewhere around (equivalent to the containing `data` only request), the second one will be included in upcoming story.
> Why there are no API v1 tests for this functionality? Will you please add them?

APIv1 is not impacted by this PR.
> Regarding tests, you can autogenerate them

Once again, will take advantage of it when working on property-based testing story.

leszke:
And since I cannot reply to the general comment above, here go my replies:
> You use both operation_id and operationId in your code can you please make it consistent across the code. I assume operation_id is more suitable.

This module takes a lot from original `CastAndValidate` plug source (I couldn't find any better way to impact its behavior only partially without copying things). I didn't want to change it simply because you can see there are no changes in how its functions work, I only wrap some additional stuff around them to enforce the behavior I expect.

> Please add test what happens if you drop both alert and service and another one where you drop all 3 (with data too).

The first scenario is already somewhere around (equivalent to the containing `data` only request), the second one will be included in upcoming story.
> Why there are no API v1 tests for this functionality? Will you please add them?

APIv1 is not impacted by this PR.
> Regarding tests, you can autogenerate them

Once again, will take advantage of it when working on property-based testing story.

lib/mongoose_push_web/plugs/cast_and_validate.ex
aleklisi: But it is just one function with multiple heads, so more than one tag for the same functions does nothing but increases lines count.

lib/mongoose_push_web/plugs/cast_and_validate.ex
aleklisi: So indlude the one needed doc line please.

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,133 @@
...
+          OpenApiSpex.Plug.Cache.adapter().put(
+            private_data.spec_module,
+            {private_data.spec, operation_lookup}
+          )
+
+          operation
+
+        operation ->
+          operation
+      end
```
aleklisi: Maybe you can extract getting operation to a function?
```elixir
    lookup_result = private_data.operation_lookup[{controller, action}]
    get_operation(lookup_result, controller, action, private_data)
    ...
      defp get_operation(nil, controller, action, private_data) do
          operationId = controller.open_api_operation(action).operationId
          operation = private_data.operation_lookup[operationId]

          operation_lookup =
            private_data.operation_lookup
            |> Map.put({controller, action}, operation)

          OpenApiSpex.Plug.Cache.adapter().put(
            private_data.spec_module,
            {private_data.spec, operation_lookup}
          )

          operation
      end
      defp get_operation(operation, _, _, _) do
          operation
      end
```

lib/mongoose_push_web/plugs/cast_and_validate.ex
aleklisi: Such a situation will never happen because the `Kernel` module is always imported so such private helper would conflict with already existing function. An attempt of defining such function will result in compilation error `imported Kernel.update_in/3 conflicts with local function`.

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
aleklisi: This change should be independent of property-based testing.

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
aleklisi: This is a small change, you can just copy-paste the code from a snippet and also it is a perfect case for using `|>` operator as shown here: https://elixirschool.com/en/lessons/basics/pipe-operator/

lib/mongoose_push_web/plugs/cast_and_validate.ex
aleklisi: Since `:operationId` should be always there it does not matter much, but it is not about writing defensive code, it is about throwing error in the right place.

lib/mongoose_push_web/plugs/cast_and_validate.ex
michalwski: For someone less experienced with Elixir, like me, I find the `Kernel.` prefix very helpful.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: Using `mix format` also does nothing but increases lines count sometimes. Let's be consistent with more experienced people [remarks](https://github.com/esl/MongoosePush/pull/166#discussion_r430577454).

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: Maybe I am repeating myself, but this code originates from openapispex repository and for the reasons I already elaborated on I decided to keep it untouched. Please head to their repo and raise an issue if you think there is any.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: Of course it is possible to overcome easily with selective imports, however these combined with eg. `Kernel` module should never pass the review, imho. My point was  that I just like to see explicitly which function is called, without scrolling back and forth to ensure there's not a single line of dirty code.

And also what Michal said, his point of view should also be considered here. It doesn't cost us so much. And keep in mind not every developer knows the whole standard library by heart.

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: And what will be the advantage of yet another helper function over case construction?

Once again, please raise the issues with code I quickly copied (for the reasons I already depicted yesterday) in [official OpenApiSpex repository](https://github.com/open-api-spex/open_api_spex) directly.

test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs
leszke: Alek, don't get me wrong, I do know what the pipe operator is. Since I misunderstood a bit my current story purpose (that actually we aim to keep already existing test scenarios), I will incorporate your suggestions as this code won't disappear right after I do that.

rslota (commented):
Hi @leszke !

I briefly looked at this code and I don't think I understand how it actually works. I can see that we copied a lot of code from the original plug - why?
The easiest solution I can think of the following:
1. Run the original request as plug-pass-through.
[success] 1.1. Return whatever original plug returned
[schema failure] 1.2. run `update_schema_and_do_call` and return it's result.
2. The End.

So yeah, in case of an schema error we call the original plug twice - once with full schema and second time with modified, smaller one. Am I missing something here? If not, this solution would be much, much shorter and easier :)

leszke:
> Hi @leszke !
> 
> I briefly looked at this code and I don't think I understand how it actually works. I can see that we copied a lot of code from the original plug - why?
> The easiest solution I can think of the following:
> 
>     1. Run the original request as plug-pass-through.
>        [success] 1.1. Return whatever original plug returned
>        [schema failure] 1.2. run `update_schema_and_do_call` and return it's result.
> 
>     2. The End.
> 
> 
> So yeah, in case of an schema error we call the original plug twice - once with full schema and second time with modified, smaller one. Am I missing something here? If not, this solution would be much, much shorter and easier :)

Hi @rslota !

I did these shenanigans mainly because I was - and still I am but not so completely now, will try to verify shortly - convinced that at the moment we exit `OpenApiSpex.Plug.CastAndValidate.call/2` function, the response is already sent to the client. What is more, I played a bit with the `conn` this function returns, and tried to act on 'schema failure' as you suggest. What can I remember from this lesson was that in the scenario that is of our interest here, we receive 422 response code. The problem is this code is also thrown by other negative scenarios in our test suites, so I thought relying just on that may not be the case.

I copied so huge amount of one `call/2` function, because in its middle (the line under `if operation.operationId do`) it calls just a `call` function without specyfing the module. If I just pointed here the original sources instead, like I do in the clauses above, It would call non-modified clause, also from original sources, and the whole plug wouldn't be executed.

rslota:
@leszke ,

> I did these shenanigans mainly because I was - and still I am but not so completely now, will try to verify shortly - convinced that at the moment we exit `OpenApiSpex.Plug.CastAndValidate.call/2` function, the response is already sent to the client.

Please verify that - I looked at their code, and all I see is call to `Plug.Conn.halt()`. This **does not** send response to the client, it only prevents other plugs from running after this one finishes, going straight to response to the client, but this should not happen if you intercept this `conn` object before returning from your plug. It's technically possible to send response immediately as side effect of the plug (`Plug.Conn.send_resp/1`), but this is very rarely done and I cannot find this in the `OpenApiSpex.Plug.CastAndValidate.call/2`.


> What is more, I played a bit with the `conn` this function returns, and tried to act on 'schema failure' as you suggest. What can I remember from this lesson was that in the scenario that is of our interest here, we receive 422 response code.

If we indeed return this error code, we do that in our controllers. That's far, far from being even executed at this stage. Plugs are executed by design sequentially and controllers are definitely (also by design) after whatever we do here. What's more, we can run this plug only if entry status code is not yet set only, which would mitigate this issue regardless of context of this plug.

Let me know what are you findings :) Also, if you're unsure what is going on in some specific case, you can link me to specific hash commit and test that I could quickly run to reproduce the issue and comment on it.

leszke:
@rslota ,

> Please verify that - I looked at their code, and all I see is call to `Plug.Conn.halt()`. This **does not** send response to the client, it only prevents other plugs from running after this one finishes, going straight to response to the client, but this should not happen if you intercept this `conn` object before returning from your plug. It's technically possible to send response immediately as side effect of the plug (`Plug.Conn.send_resp/1`), but this is very rarely done and I cannot find this in the `OpenApiSpex.Plug.CastAndValidate.call/2`.

You were right and it doesn't send any reponse itself actually.

> If we indeed return this error code, we do that in our controllers. That's far, far from being even executed at this stage. Plugs are executed by design sequentially and controllers are definitely (also by design) after whatever we do here. What's more, we can run this plug only if entry status code is not yet set only, which would mitigate this issue regardless of context of this plug.

Using your courtesy, I updated the code to give your approach a try, and what's good is that our tests are working correctly, ie. I almost got rid (at one moment completely) of this 'no schema provided' issue, also when some weird changes has been made in deeper parts of the request. What is bad, and I'm afraid is actually revealing some issue, is that manual, ie. `curl` requests are not working, even when their functional-tests-equivalents are. Could you please have a look here `eb1bea12` and try to hint something?
Steps to reproduce:
`mix test test/mongoose_push_web/controllers/api_v3_notification_controller_test.exs:46`
and its manual equivalent
`curl --header "Content-Type: application/json" --data '{"service" : "apns","alert" : {"title" : "title", "body" :""}, "priority": "incorrect value"}' https://localhost:8443/v3/notification/test -k`
After second try your phoenix server terminal should be flooded with messages like `text[91][100][27]=111 text[91][100][28]=47 text[91][100][29]=77`.

rslota:
@leszke ,

Actually it seems you were right - this plug does send the response right away, but it's buried quite deeply, and I think this is the reason of your errors.
I've pushed 2 commits to my fork:
1. Fix for loge messages that are iodata() based: https://github.com/esl/MongoosePush/commit/3fef193616f9b15d2bf032ea0a6cef8d664fda07 (highly recommend you cherry pick that ;)
2. Proposal changes that prevent this plug from sending the response: https://github.com/esl/MongoosePush/commit/f6454e75924450013368233847bc1c9578dd607e

Yes, this is a hack, but this whole solution is a hack, so to me this is just a question of whether we want short or a long one :)
Essentially, my commit replaces HTTP adapter that is used normally (Cowboy) with not even complete implementation of my own that does literally nothing. This means we can use first call to OpenApiSpex to just get the error code, and subsequent one to actually do stuff with modified or not schema.

rslota:
>     2\. Proposal changes that prevent this plug from sending the response: [f6454e7](https://github.com/esl/MongoosePush/commit/f6454e75924450013368233847bc1c9578dd607e)

Obviously my commit is very quick and dirty - if you decide to go this route please define the stub adapter in separate file and make it conform to the behavior fully, not just the function that happens to be called in this case ;)

leszke:
Hi again @rslota,

I applied your proposal and looks like it works :) I also gave my best to implement the stubbed adapter the way that both conforms the behaviour and still makes any sense. Thank you for your commitment :)

lib/mongoose_push_web/plugs/cast_and_validate.ex
```diff
@@ -0,0 +1,80 @@
...
+  @impl Plug
+  def init(opts) do
+    OpenApiSpex.Plug.CastAndValidate.init(opts)
+  end
+
+  @impl Plug
+  def call(conn, opts) do
+    stub_conn = %Plug.Conn{
+      conn
+      | adapter: {MongoosePushWeb.Plug.CastAndValidateStubAdapter, %{}}
```
rslota: How about:
```suggestion
      | adapter: {MongoosePushWeb.Plug.CastAndValidate.StubAdapter, %{}}
```
with the file rename as well ;)

rslota (commented):
@leszke ,

This solution to me looks good, I would only rename the adapter (see comment), otherwise I have nothing more to add. Great job!
I haven't do any "code review", I only looked at the "solution method" - because of that I won't "approve" this PR, this is simply not my call, so please make sure other reviewers see the code :)

lib/mongoose_push_web/plugs/cast_and_validate.ex
leszke: @rslota,

You got me, I just was too lazy to create a separate folder for this file, what I will make up in the upcoming PR. Thanks again for your engagement, obviously I will wait for guys to approve this before we merge :)

michalwski (approved):

michalwski:
@rslota thanks for keeping an eye on the changes! Much appreciated!

PR #169, opened 2020-06-15
Metrics - refactoring notification send time

jan-ciesla:
This PR updates:
-  Metric definition with updated measurement units. 'second' was inconsistent with the histogram buckets and all events were smaller than the lowest bucket.
- Event name is overridden with prometheus recommended format that includes the unit of the measurement in the event name. 

Previous integration tests are checking this functionality, hence no need for additional tests.

### DIFF
--- a/lib/mongoose_push/metrics/telemetry_metrics.ex
+++ b/lib/mongoose_push/metrics/telemetry_metrics.ex
@@ -10,10 +10,12 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
   def metrics do
     [
       # Summary is not yet supported in TelemetryMetricsPrometheus
-      Telemetry.Metrics.distribution("mongoose_push.notification.send.time",
-        buckets: [100, 250, 500, 1000],
-        tags: [:status, :service, :error_category, :error_reason],
-        unit: {:native, :second}
+      Telemetry.Metrics.distribution(
+        "mongoose_push.notification.send.time.microsecond",
+        event_name: [:mongoose_push, :notification, :send],
+        measurement: :time,
+        buckets: [1000, 10_000, 25_000, 50_000, 100_000, 250_000, 500_000, 1000_000],
+        tags: [:status, :service, :error_category, :error_reason]
       ),
 
       # measurement is ignored in Counter metric
--- a/test/unit/mongoose_push_telemetry_metrics_test.exs
+++ b/test/unit/mongoose_push_telemetry_metrics_test.exs
@@ -38,13 +38,13 @@ defmodule MongoosePushTelemetryMetricsTest do
 
     # Distribution metric contains count as well as the buckets
     fcm_regex =
-      ~r/mongoose_push_notification_send_time_count{error_category=\"\",error_reason=\"\",service=\"fcm\",status=\"success\"} (?<count>[\d]+)/
+      ~r/mongoose_push_notification_send_time_microsecond_count{error_category=\"\",error_reason=\"\",service=\"fcm\",status=\"success\"} (?<count>[\d]+)/
 
     fcm_match = Regex.named_captures(fcm_regex, metrics)
     fcm_count = get_count(fcm_match)
 
     apns_regex =
-      ~r/mongoose_push_notification_send_time_count{error_category=\"\",error_reason=\"\",service=\"apns\",status=\"success\"} (?<count>[\d]+)/
+      ~r/mongoose_push_notification_send_time_microsecond_count{error_category=\"\",error_reason=\"\",service=\"apns\",status=\"success\"} (?<count>[\d]+)/
 
     apns_match = Regex.named_captures(apns_regex, metrics)
     apns_count = get_count(apns_match)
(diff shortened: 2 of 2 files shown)


lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -10,10 +10,13 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
   def metrics do
     [
       # Summary is not yet supported in TelemetryMetricsPrometheus
-      Telemetry.Metrics.distribution("mongoose_push.notification.send.time",
-        buckets: [100, 250, 500, 1000],
+      Telemetry.Metrics.distribution(
+        "mongoose_push.notification.send.time.microsecond",
```
rslota, line 14: I'm very glad to see that change :)

lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -10,10 +10,13 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
...
+        event_name: [:mongoose_push, :notification, :send],
+        measurement: :time,
+        buckets: [10, 25, 50, 100, 200, 500, 1000],
```
rslota: Those buckets seem to be for milliseconds (at least they make sense for milliseconds), while the change is for `microsecond`. Is that a mistake or am I missing something?

lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -10,10 +10,13 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
...
         tags: [:status, :service, :error_category, :error_reason],
-        unit: {:native, :second}
+        unit: {:native, :microsecond}
```
rslota: `second` is recommended default for Prometheus in general. Is there a reason for using non-default unit?

michalwski: In all our installations (MongooseIM + MongoosePush) metrics showing a dalay of execution time of something were in microseconds. So the conistency is the reason here.

rslota: Too bad, but fair enough ;)

lib/mongoose_push/metrics/telemetry_metrics.ex
jan-ciesla: When doing local testing setting the buckets in this manner resulted in having hits for some buckets. The buckets were extended to have results granularity when testing with mock and fcm.

jan-ciesla: Do you have some intuition as to what order of magnitude those results should be for installation with some significant load?

lib/mongoose_push/metrics/telemetry_metrics.ex
jan-ciesla: When setting seconds we'd need to adjust the bucket sizes to be really small. This way the results are more readable.

lib/mongoose_push/metrics/telemetry_metrics.ex
rslota: Those values, in `microsecond` mean that the max bucket is for 1ms. Indeed, on local environment, where mocks are on "localhost", this is possible. I don't have any actual data from production environments on that, but APNS/FCM server are always remote, so the request latency itself will be **at least** few milliseconds. I wouldn't be surprised if the whole process of sending notification would be more then 200ms, so 200x max bucket right now. That's why I said, that current buckets look like values in `milliseconds` (with last value at 1000ms == 1sec).

lib/mongoose_push/metrics/telemetry_metrics.ex
rslota: @janciesla8818 This doesn't matter at all, this is just a unit. It doesn't matter if buckets are `[10, 25, 50, 100, 200, 500, 1000]` milliseconds or `[0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1.0]` seconds. Those are the same. On Grafana or whatever you use for displaying this, you choose display unit separately anyway. Here we only choose unit in which this will be saved to Prometheus and which needs to be selected in Grafana as "source unit", nothing else.
That's why "conistency" reason with other service makes sense, but functionally there is no difference.

michalwski: BTW, do we need to make the conversion from `native` to `microsecond`. As far as I can tell the event `[:mongoose_push, :notification, :send]` reports in `microsecond` already.

lib/mongoose_push/metrics/telemetry_metrics.ex
michalwski: Ok, we are going to adjust the buckets.

michalwski (commented):
Thanks for the change @janciesla8818.

Thanks for your input @rslota. It's good to see you are still interested in helping us with MongoosePush!

lib/mongoose_push/metrics/telemetry_metrics.ex
rslota: IMHO, the measurements should be made in `native` (if it isn't I wound change that) and if that happens to be the same as `microsecond`, it doesn't matter, having explicit translation here is preferable.

michalwski: @janciesla8818  @rslota In case of this event, it's done with `:timer.tc` which returns the time in `microsecond` (http://erlang.org/doc/man/timer.html#tc-3). The `native` time unit can be different though. On my machine, it's in `nanosecond`.
I have a hunch that what happens now is converting a value given in **microsecond** from **_native_** to **_microsecond_** will give the wrong result (at least on those machines where `native` equals to `nanosecond`)

rslota: @michalwski @janciesla8818 ,
Indeed, `:timer.tc` changes time unit explicitly to `microsecond`. What I normally see in projects is custom macro `timed` that is essentially a copy of `:timer.tc`, but without time conversion from native and without additional unnamed function (for better stack-traces). This has added benefit of having in project "the" function/macro fir timing, no need to guess which standard one could be used.
This is what I would do ^, but obviously if you plan on using `:timer.tc`, then you need to make sure you only use that, and remove then conversion for metrics, or leave `{:microsecond , :microsecond}` for explicit context.

jan-ciesla: @rslota , @michalwski thanks for the comments. I'll have it removed. We'll have the unit in the name of the event, so putting additional `{:microsecond , :microsecond}` could be a bit redundant.

michalwski (approved):

PR #171, opened 2020-06-16
Metrics documentation update

michalwski:
This PR is primarily about metrics documentation update. It also contains the following changes which may be extracted to separate PR based on review comments.

* Start the `MongoosePush.Metrics.TelemetryMetrics` child befor others to captuer events at startup
* Short metrics documentation in the code
* Add example Prometheus config for MongoosePush and docker-compose file to help with a basic setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,11 +517,13 @@ Where:
 * `SERVICE` is either `"apns"` or `"fcm"`
 * `CATEGORY` is an arbitrary error category term (in case of `status="error"`) or an empty string (when `status="success"`)
 * `REASON` is an arbitrary error reason term (in case of `status="error"`) or an empty string (when `status="success"`)
-* `LE` defines the `less or equal` values for buckets, currently `1000`, `10_000`, `25_000`, `50_000`, `100_000`, `250_000`, `500_000`, `1000_000` or `+Inf`
+* `LE` defines the `upper inclusive bound` (`less than or equal`) values for buckets, currently `1000`, `10_000`, `25_000`, `50_000`, `100_000`, `250_000`, `500_000`, `1000_000` or `+Inf`
 
 > **NOTE**
-> A mesurment of value 50_000 will be added to all buckets which are less or equal to 50_000.
-> In this case these are buckets `50_000`, `100_000`, `250_000`, `500_000`, `1000_000` and `+Inf`
+>
+> A bucket of value 250_000 will keep the count of measurements that are less than or equal to 250_000.
+> A measurement of value 51_836 will be added to all the buckets where the upper bound is greater than 51_836.
+> In this case these are buckets `100_000`, `250_000`, `500_000`, `1000_000` and `+Inf`
 
 This histogram metric shows the distribution of times needed to:
 1. Select a worker (this may include waiting time when all workers are busy).
@@ -530,12 +532,12 @@ This histogram metric shows the distribution of times needed to:
 
 ##### Counters
 
-* `mongoose_push_supervisor_init_count{service=${SERVICE}}` - Counts number of push notification service supervisor starts.
+* `mongoose_push_supervisor_init_count{service=${SERVICE}}` - Counts the number of push notification service supervisor starts.
   The `SERVICE` variable can take `"apns"` or `"fcm"` as a value.
-  This metrics is update when MongoosePush starts or later when the underlying supervision tree is terminated and the error is propagate to the main application supervisor.
-* `mongoose_push_apns_state_init_count` - Counts number of APNS state initialisations.
-* `mongoose_push_apns_state_terminate_count` - Counts number of APNS state terminations.
-* `mongoose_push_apns_state_get_default_topic_count` - Counts number of default topic reads from cache.
+  This metric is updated when MongoosePush starts, and later on when the underlying supervision tree is terminated and the error is propagated to the main application supervisor.
+* `mongoose_push_apns_state_init_count` - Counts the number of APNS state initialisations.
+* `mongoose_push_apns_state_terminate_count` - Counts the number of APNS state terminations.
+* `mongoose_push_apns_state_get_default_topic_count` - Counts the number of default topic reads from cache.
 
 #### How to quickly see all metrics
 
@@ -544,7 +546,7 @@ curl -k https://127.0.0.1:8443/metrics
 ```
 
 The above command assumes that MongoosePush runs on `localhost` and listens on port `8443`.
-Please, mind the `HTTPS` protocol, metrics are hosted on the same port as other API.
+Please, mind the `HTTPS` protocol, metrics are hosted on the same port than all the other API endpoints.
 
 #### Prometheus configuration
 
@@ -564,5 +566,3 @@ scrape_configs:
           group: 'production'
 
 ```
-
-
--- a/README.md
+++ b/README.md
@@ -491,26 +491,40 @@ If you specify both **alert** and **data**, target device will receive both noti
 * **500** `{"reason" : reason}` - the server internal error occured,
   specified by **reason**.
 
-### I use MongoosePush docker, where do I find `sys.config`?
+### Metrics
 
-If you use dockerized MongoosePush, you need to do the following:
-* Start MongoosePush docker, let's assume its name is `mongoose_push`
-* Run: `docker cp mongoose_push:/opt/app/var/sys.config sys.config` on you docker host (this will get the current `sys.config` to your `${CWD}`)
-* Modify the `sys.config` as you see fit (for metrics, see above)
-* Stop MongoosePush docker container and restart it with the modified `sys.config` as volume in `/opt/app/sys.config` (yes, this is not the path we used to copy this file from, this is an override)
+MongoosePush 2.1 provides metrics in the Prometheus format on the `/metrics` endpoint.
+This is a breaking change compared to previous releases.
+Existing dashboards will need to be updated.
+
+#### Available metrics
+
+#### How to quickly see all metrics
+
+```bash
+curl -k https://127.0.0.1:8443/metrics
+```
+
+The above command assumes that MongoosePush runs on `localhost` and listens on port `8443`.
+Please, mind the `HTTPS` protocol, metrics are hosted on the same port as other API.
+
+#### Prometheus configuration
+
+When configuring Prometheus, it's important to:
+* set the `scheme` to `https` since MongoosePush exposes `/metrics` path encrypted endpoint (HTTPS)
+* set the `insecure_skip_verify` to `true` if the default self-signed certificates are used
+
+```yaml
+scrape_configs:
+  - job_name: 'mongoose-push'
+    scheme: 'https' #MongoosePush exposes encrypted endpoint - HTTPS
+    tls_config: #The default certs used by MongoosePush are self-signed
+      insecure_skip_verify: true #For checking purposes we can ignore certs verification
+    static_configs:
+      - targets: ['mongoose-push:8443']
+        labels:
+          group: 'production'
+
+```
 
 
-### Available metrics
-
-The following metrics are available:
-
-* `mongoose_push_apns_state_get_default_topic_count`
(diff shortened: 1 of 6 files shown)


lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -15,17 +15,28 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
         event_name: [:mongoose_push, :notification, :send],
         measurement: :time,
         buckets: [1000, 10_000, 25_000, 50_000, 100_000, 250_000, 500_000, 1000_000],
-        tags: [:status, :service, :error_category, :error_reason]
+        tags: [:status, :service, :error_category, :error_reason],
+        description:
+          "A histogram showing push notification sent times. Includes worker selection (with possible waiting if all are busy)"
```
NelsonVides: Great addition! 😄

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
-* **STATUS** is either `success` or `error`
-* **LENGTH** is either `100` or `250` or `500` or `1000` or `+Inf`
+* `STATUS` is `"success"` for the successful notifications or `"error"` in all other cases
+* `SERVICE` is either `"apns"` or `"fcm"`
+* `CATEGORY` is an arbitrary error category term (in case of `status="error"`) or an empty string (when `status="success"`)
+* `REASON` is an arbitrary error reason term (in case of `status="error"`) or an empty string (when `status="success"`)
+* `LE` defines the `less or equal` values for buckets, currently `1000`, `10_000`, `25_000`, `50_000`, `100_000`, `250_000`, `500_000`, `1000_000` or `+Inf`
+
+> **NOTE**
+> A mesurment of value 50_000 will be added to all buckets which are less or equal to 50_000.
```
NelsonVides:
```suggestion
> A measurement of value 50_000 will be added to all buckets which are less or equal to 50_000.
```

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+> In this case these are buckets `50_000`, `100_000`, `250_000`, `500_000`, `1000_000` and `+Inf`
+
+This histogram metric shows the distribution of times needed to:
+1. Select a worker (this may include waiting time when all workers are busy).
+2. Send a request.
+3. Get a response from push notifications provider.
+
+##### Counters
+
+* `mongoose_push_supervisor_init_count{service=${SERVICE}}` - Counts number of push notification service supervisor starts.
```
NelsonVides:
```suggestion
* `mongoose_push_supervisor_init_count{service=${SERVICE}}` - Counts the number of push notification service supervisor starts.
```

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+  The `SERVICE` variable can take `"apns"` or `"fcm"` as a value.
+  This metrics is update when MongoosePush starts or later when the underlying supervision tree is terminated and the error is propagate to the main application supervisor.
+* `mongoose_push_apns_state_init_count` - Counts number of APNS state initialisations.
```
NelsonVides:
```suggestion
* `mongoose_push_apns_state_init_count` - Counts the number of APNS state initialisations.
```

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+* `mongoose_push_apns_state_terminate_count` - Counts number of APNS state terminations.
```
NelsonVides:
```suggestion
* `mongoose_push_apns_state_terminate_count` - Counts the number of APNS state terminations.
```

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+* `mongoose_push_apns_state_get_default_topic_count` - Counts number of default topic reads from cache.
```
NelsonVides:
```suggestion
* `mongoose_push_apns_state_get_default_topic_count` - Counts the number of default topic reads from cache.
```

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+This histogram metric shows the distribution of times needed to:
+1. Select a worker (this may include waiting time when all workers are busy).
+2. Send a request.
+3. Get a response from push notifications provider.
+
+##### Counters
+
+* `mongoose_push_supervisor_init_count{service=${SERVICE}}` - Counts number of push notification service supervisor starts.
+  The `SERVICE` variable can take `"apns"` or `"fcm"` as a value.
+  This metrics is update when MongoosePush starts or later when the underlying supervision tree is terminated and the error is propagate to the main application supervisor.
```
NelsonVides:
```suggestion
  This metrics is updated when MongoosePush starts, and later on when the underlying supervision tree is terminated and the error is propagated to the main application supervisor.
```

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+* `mongoose_push_apns_state_get_default_topic_count` - Counts number of default topic reads from cache.
+
+#### How to quickly see all metrics
+
+```bash
+curl -k https://127.0.0.1:8443/metrics
+```
+
+The above command assumes that MongoosePush runs on `localhost` and listens on port `8443`.
+Please, mind the `HTTPS` protocol, metrics are hosted on the same port as other API.
```
NelsonVides:
```suggestion
Please, mind the `HTTPS` protocol, metrics are hosted on the same port than all the other API endpoints.
```
Or something like that, I'm not sure. I'd treat "API" as an indefinite singular noun, so "as other API" doesn't sound grammatically correct 🤔

README.md
```diff
@@ -491,26 +491,78 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+    tls_config: #The default certs used by MongoosePush are self-signed
+      insecure_skip_verify: true #For checking purposes we can ignore certs verification
+    static_configs:
+      - targets: ['mongoose-push:8443']
+        labels:
+          group: 'production'
+
+```
+
+
```
NelsonVides: Not needed empty lines
```suggestion
```

lib/mongoose_push/application.ex
```diff
@@ -33,8 +33,10 @@ defmodule MongoosePush.Application do
     _ = check_runtime_configuration_status()
 
     # Define workers and child supervisors to be supervised
+    # The MongoosePush.Metrics.TelemetryMetrics child is started first to capture possible events
+    # when services start
     children =
-      service_children() ++ [MongoosePushWeb.Endpoint, MongoosePush.Metrics.TelemetryMetrics]
+      [MongoosePush.Metrics.TelemetryMetrics] ++ service_children() ++ [MongoosePushWeb.Endpoint]
```
NelsonVides, line 39: Good catch 🙂

lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -15,17 +15,28 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
...
         tags: [:error_reason],
-        tag_values: fn metadata -> %{metadata | error_reason: metadata.reason} end
+        tag_values: fn metadata -> %{metadata | error_reason: metadata.reason} end,
+        description: "Counts number of APNS state terminations"
       ),
-      Telemetry.Metrics.counter("mongoose_push.apns.state.get_default_topic.count")
+      Telemetry.Metrics.counter(
+        "mongoose_push.apns.state.get_default_topic.count",
+        description: "Counts number of APNS default topic reads from the ETS cache"
+      )
```
NelsonVides: These are also missing the `the` articles, as in "Counts _the_ number of..."

NelsonVides (commented):
Fantastic!

The only thing I'd say is that MPush docs, i.e., a single README, is getting out of control with the size, but reorganising that is on a different task 🙂 

Oh, and some articles were missing 😛

README.md
kmakiela:
```suggestion
  This metric is updated when MongoosePush starts or later when the underlying supervision tree is terminated and the error is propagated to the main application supervisor.
```

test/docker/docker-compose.mpush.yml
```diff
@@ -1,5 +1,5 @@
 # This file needs to be used along with `docker-compose.mocks.yml`:
-# docker-compose -f test/docker/docker-compose.mocks.yml -f test/docker/docker-compose.mpush.yml ...
+# PRIV=priv docker-compose -f test/docker/docker-compose.mocks.yml -f test/docker/docker-compose.mpush.yml ...
```
kmakiela, line 2: Why do we need this change?

kmakiela (commented):
Great update! However, reading through 600 lines of readme is a bit much so that may be something we want to change in the near future.

test/docker/docker-compose.mpush.yml
michalwski: The `docker-compose-mpush.yml` requires the `PRIV` env var. It's set for us when we do `mix text.env.up` but when running from console we need to export it manually.

michalwski:
@NelsonVides @kmakiela thanks for the review. I 100% agree with the documentation split/refactoring. This is definitely a subject for a separate PR.

README.md
jan-ciesla: Not sure if I understand this right, I think this is the other way around. The bucket contains measurements that are less or equal to the bucket size.

`More particularly they're counters which form a cumulative histogram, le stands for less than or equal to. So 26688 requests took less than or equal to 200ms, 27760 requests took less than or equal to 400ms, and there were 28860 requests in total. ` 
[https://www.robustperception.io/how-does-a-prometheus-histogram-work](https://www.robustperception.io/how-does-a-prometheus-histogram-work)

README.md
jan-ciesla:
```suggestion
Please, mind the `HTTPS` protocol, metrics are hosted on the same port as other API endpoints.
```
How about this as compromise? 
`the same as` seems more grammatically correct than `the same than`

lib/mongoose_push/metrics/telemetry_metrics.ex
jan-ciesla: `sent` or `send`? I always confuse when to use the two 😄 , but the event name has `send.time`

are they both correct or should we change one?

lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -15,17 +15,28 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
...
+      ),
       Telemetry.Metrics.counter("mongoose_push.apns.state.terminate.count",
         tags: [:error_reason],
-        tag_values: fn metadata -> %{metadata | error_reason: metadata.reason} end
+        tag_values: fn metadata -> %{metadata | error_reason: metadata.reason} end,
+        description: "Counts number of APNS state terminations"
       ),
-      Telemetry.Metrics.counter("mongoose_push.apns.state.get_default_topic.count")
+      Telemetry.Metrics.counter(
+        "mongoose_push.apns.state.get_default_topic.count",
```
jan-ciesla:
```suggestion
      Telemetry.Metrics.counter("mongoose_push.apns.state.get_default_topic.count",
```

just for consistency with the remaining events.

jan-ciesla (changes requested):
Good stuff! I really like the PR, we just need to make sure that the part regarding `less or equal` is accurate as it can be a bit confusing.

lib/mongoose_push/metrics/telemetry_metrics.ex
michalwski: That's probably `send` better to be consistent with the event and metric name.

README.md
michalwski: Right, I messed with that. Good catch!

michalwski:
@janciesla8818 thanks for the review. I think I applied all the comments now.

jan-ciesla (approved):

NelsonVides (approved):

PR #172, opened 2020-06-16
Openapi property based testing

leszke:
This PR introduces property-based approach to APIv3 testing, with a little help of `StreamData` framework. I also incorporated the fix/enhancement of 'the hack of the hack' introduced in #168. Shortly speaking, the request containing, from the needed fields, only `data`, `alert/body` and `alert/title` fields, was matched against `SilentNotification` schema, what has been leading to improper `Unexpected field: alert` error.

I also improved `ControllersHelper.missing_field_response` helper function to better reflect possible error messages we can encounter during tests, basing on the API version being used.


test/support/requests_generator.ex
```diff
@@ -0,0 +1,77 @@
...
+    nonempty_string()
+  end
+
+  def mandatory_field() do
+    one_of_strings([:service, :title, :body])
+  end
+
+  # basic types
+  defp nonempty_string() do
+    StreamData.string(:alphanumeric) |> StreamData.filter(fn x -> String.length(x) > 0 end)
```
NelsonVides: What about something like (or whatever the correct syntax is)
```suggestion
    StreamData.string(:alphanumeric, min_length: 1)
```
Note that `StreamData.string/2` can take the same options than `list_of/1`.

DenysGonchar:
looks good. please recheck Nelson's suggestion.

test/support/requests_generator.ex
leszke: good catch 😀

NelsonVides (approved):
Great work! 😄

PR #174, opened 2020-06-17
MongoosePush JSON log formatter

zofpolkowska:
This PR introduces configurable JSON log formatter. By default for `dev` and `test` environments the format is FMT, for `production` - JSON. It can be configured by sevia toml `general.logging.format` key (in TOML schema) or `PUSH_LOGFORMAT` env variable to `logfmt`/`json`.


lib/mongoose_push/logger/json.ex
```diff
@@ -0,0 +1,41 @@
...
+    function = Keyword.get(metadata, :function)
+    line = Keyword.get(metadata, :line)
+    pid = Keyword.get(metadata, :pid)
+    what = Keyword.get(metadata, :what)
+
+    custom_metadata =
+      metadata
+      |> Keyword.drop([:module, :function, :line, :pid, :mfa, :time, :gl, :what])
+      |> flatten_metadata()
+      |> Enum.reduce(%{}, fn {key, val}, acc -> Map.put(acc, key, val) end)
```
leszke: `|> Enum.into(Map.new())` or even `|> Enum.into(%{})`? 🙂

leszke (commented):
Thanks for the PR! As I'm currently playing with k8s stuff before tomorrow, I only went over the PR and pointed out one minor call that imo can be rewritten for better reading experience 🙂

lib/mongoose_push/application.ex
```diff
@@ -159,6 +161,14 @@ defmodule MongoosePush.Application do
     end
   end
 
+  defp set_logformat(:fmt), do: set_loglevel(MongoosePush.Logger.Fmt)
+
+  defp set_logformat(:json), do: set_loglevel(MongoosePush.Logger.JSON)
```
NelsonVides: Are these calls correct? I'm wondering about the fact that `set_logformat` calls `set_loglevel` with a module, not sure if it's some Elixir magic, but it doesn't look intuitive to call later on `Logger.configure(level: MongoosePush.Logger.Fmt)` 🤔

lib/mongoose_push/logger/fmt.ex
```diff
@@ -0,0 +1,39 @@
+defmodule MongoosePush.Logger.Fmt do
+  @moduledoc """
+  Module responsible for formatting logger output
```
NelsonVides: This moduledoc could be updated to reflect that is only doing a specific kind of formatting now 🙂

lib/mongoose_push/logger/json.ex
```diff
@@ -0,0 +1,41 @@
+defmodule MongoosePush.Logger.JSON do
+  @moduledoc """
+  JSON log formatter
```
NelsonVides: Same here about that doc

NelsonVides (commented):
Nice move for the common functionality in Logger.Common, and I also learned some things about Elixir Logger here 🙂 
What about tests to check that we're effectively formatting the logs one way or another? Was there any such test for fmt before? Could there be specialised tests for both formats?

lib/mongoose_push/application.ex
zofpolkowska: Thanks! Yeah, tests indeed would be nice, there weren't any for the Fmt, but I'll think of a way to test both formatters now.

config/config.exs
```diff
@@ -7,7 +7,7 @@ config :logger,
   handle_sasl_reports: false
 
 config :logger, :console,
-  format: {MongoosePush.LoggerFmt, :format},
+  format: {MongoosePush.Logger.Fmt, :format},
```
rslota: Technically the format is called: `logfmt`, not `fmt`. I would suggest to change this name throughout to be more accurate - `fmt` alone feels weird :)

config/prod.exs
```diff
@@ -32,7 +32,9 @@ config :mongoose_push, MongoosePushWeb.Endpoint,
   check_origin: true,
   server: true
 
-config :mongoose_push, :logging, level: {:system, :atom, "PUSH_LOGLEVEL", :info}
+config :mongoose_push, :logging,
+  level: {:system, :atom, "PUSH_LOGLEVEL", :info},
+  format: {:system, :atom, "PUSH_LOGFORMAT", :logjson}
```
rslota: Actually `json` is still `json` :) Sorry for confusion, but I meant that `logfmt` is a name of format, in the same as `JSON` is a name of the format. There is no such thing as `logjson` or `fmt`. The `log` part of `logfmt` is just integral part of the name, not anything generic ;)

zofpolkowska: Oh, okay, thanks! :)

lib/mongoose_push/logger/json.ex
```diff
@@ -0,0 +1,41 @@
+defmodule MongoosePush.Logger.JSON do
```
michalwski, line 1: I just wonder why we create new JSON formater if there is several available. I may be missing something.

michalwski (commented):
Thanks for the change. The code looks clean and now we can finally configure the JSON format for logs. This will make MongoosePush more friendly towards different logging aggregators. 

I had just one question regarding the `MongoosePush.Logger.Json` module.

lib/mongoose_push/logger/json.ex
zofpolkowska: There are some json _backends_ for logger (like [logger_json](https://github.com/Nebo15/logger_json)). If we already have a formatter module for the `loggerfmt` I thought this would work better structure-wise. But we can use this library instead.

lib/mongoose_push/application.ex
zofpolkowska: I would like to use `ExUnit.CaptureLog` to test the formats, but there was an issue with with that which caused the tests to fail. There were several fixes for the issues resulting in similar error, last one released only in `1.10.2`. I'm working on that, just messed something up on the way :D

lib/mongoose_push/logger/json.ex
rslota: I'm honestly unsure about this library. It's not a `formatter`, it's a `backend`. I skimmed through their code quickly and I don't understand why they want this way. This is concerning to me, as backeds have to deal with IO, log throttling etc. and even if their implementation is "OK", I imagine that `console` backend that is part of Elixir standard lib is better maintained, and more importantly, better tested. They also have formatter, yes, but I'm not sure if it's compatible with `console` backend, and it's very, very simple one anyway - I'm not a fan of including external dependencies to avoid writing few lines of trivial code ;)
wdyt @michalwski @zofpolkowska ?

lib/mongoose_push/application.ex
rslota: IMHO there is very little reason to use `ExUnit.CaptureLog`, especially that it's hard to use considering that it only works for "calling process", and this application has a lot of GenServes just running somewhere and doing stuff. Also, we don't need to test Logger's `console` backend - we can trust Elixir devs here :) If we write only formatter, we know what the contract between Logger's `console` backend and the formatter is, so unit tests for the formatter IMHO should be enough.

zofpolkowska: I'll that, thanks:)

lib/mongoose_push/logger/json.ex
zofpolkowska: Yeah, with that feeling I did it the way I did it. Didn't understand the point of having backend as a formatter:)

rslota: @zofpolkowska Elixir's Logger have pluggable logging `backends` - those are `gen_event` based implementations that are suppose to handle all `Logger.log` calls in some way (write them to IO-device somehow). They need to throttle if IO is too slow, drop logs, do quite a few things to keep the system running properly when there is too much logs compared to IO performance. For example, there may be a file-based `backend` that writes to files. The one default `backend` is called `console` and it's just printing logs to, well, console. custom `formatter` is a configuration option for `console` backend specifically, this is not a global option for `Logger` itself. The `console` backend allows adding custom `formatters` that format data before it's printed on stdout.
Because of this architecture, with assumption that you don't need other backends then `console` (and I think you don't, considering Docker prefers stdout-logging), you can easily get away with just providing custom `formatter`, without need to code the entire logging `backend`. `logger_json` library however implements entire `backend` (also with support for `formatters` as they have two different JSON formatters, one "plain", second compatible with Google Cloud). I don't know why they did that, maybe they wanted generic IO-device to also log to files and still keep formatters. It's important however to make this distinction, because writing a `formatter` to existing `backend` is trivial and it's hard to break anything. Writing entire `backend` however, well, that's totally different story. 

I hope this clarifies things ;)

leszke (approved):

jan-ciesla (approved):

michalwski (approved):

PR #175, opened 2020-06-17
Metrics based on telemetry events generated by sparrow

jan-ciesla:
This PR adds new metrics based on sparrow telemetry events. The changes include: 
* Adding histogram, sum and counter metric for HTTP requests
* Count of worker pools
* Count of workers per worker pool
* Total count of workers in all worker pools

Last three metrics are gathered using `:wpool.stats` to get the information from worker pool. 

`Telemetry poller` is added to the project to collect periodic metrics. Counters of pools and workers are collected as events generated during telemetry poller events. The period for gathering metrics is 5 seconds which is the default value for telemetry poller.


lib/mongoose_push/metrics/telemetry_metrics.ex
```diff
@@ -7,6 +7,10 @@ defmodule MongoosePush.Metrics.TelemetryMetrics do
     TelemetryMetricsPrometheus.Core.child_spec(metrics: metrics())
   end
 
+  def pooler do
+    [{:telemetry_poller, measurements: periodic_measurements(), period: 5_000}]
```
michalwski: Would it make sense to allow configuring the `period`? Prometheus by default scraps metrics every minute so perhaps we don't need to calculate these stats every 5s.

README.md
```diff
@@ -519,17 +519,26 @@ Where:
 * `REASON` is an arbitrary error reason term (in case of `status="error"`) or an empty string (when `status="success"`)
 * `LE` defines the `upper inclusive bound` (`less than or equal`) values for buckets, currently `1000`, `10_000`, `25_000`, `50_000`, `100_000`, `250_000`, `500_000`, `1000_000` or `+Inf`
 
+This histogram metric shows the distribution of times needed to:
+1. Select a worker (this may include waiting time when all workers are busy).
+2. Send a request.
+3. Get a response from push notifications provider.
+
+###### HTTP/2 requests
```
michalwski, line 527: Does this metric show time spent only on sending the request and waiting for the response?

mix.exs
```diff
@@ -61,6 +61,7 @@ defmodule MongoosePush.Mixfile do
       {:telemetry, "~>0.4.1"},
       {:telemetry_metrics, "~> 0.5"},
       {:telemetry_metrics_prometheus_core, "~> 0.4"},
+      {:telemetry_poller, "~> 0.4"},
```
michalwski: Can we use version `0.5`? It seems to be the newest on hex.pm.

michalwski (commented):
Thanks for the PR! It's good to have metrics showing how much time was spent in the HTTP/2 worker. Metrics showing worker pool statistics are also valuable.

On a side note, I think it'd be worth checking if we use the newset versions of the `Telemetry` releated libraries.

mix.exs
jan-ciesla: Will have that updated.

lib/mongoose_push/metrics/telemetry_metrics.ex
jan-ciesla: Will have it changed to 30 seconds.

README.md
jan-ciesla: This metric shows only the time it takes to handle and send request. It only opens the stream and sends the request. It does not measure response time. In general, when this timer is small it it possible that we are getting errors with opening the stream.

michalwski: Thanks for checking this. I think it'd be good to add this info to the doc.

jan-ciesla: Ok, will add that info.

jan-ciesla: @michalwski, this part is already added to the doc..

michalwski: Hmm, I wonder how did you come to the conclusion that it measures only the time it takes to handle and send the request. If I'm correct this metric measures the time of this function in sparrow: https://github.com/esl/sparrow/blob/master/lib/sparrow/h2_worker.ex#L352-L430

Inside the function, we are waiting for the response and emit additional events depending on the response.

Also, I think the name of the metric starts with `sparrow_h2_worker` instead of `sparrow_h_worker`.

michalwski: All right, we discussed this offline with @janciesla8818. I was misled by the `post_result` var name and the case checking if this is a successful response or not. In fact, this var gives us information if the request was sent successfully or not. The response to the request arrives asynchronously (HTTP/2). Thanks to that we can send many requests using the same `h2_worker` and connection.

When it comes to the metric name, the `Prometheus.Core` lib removes all numbers from the metric name.

michalwski (approved):

PR #176, opened 2020-06-18
Update telemetry dependencies

NelsonVides:
```
TelemetryMetricsPrometheusCore fixed now the dialyzer spec that we
needed to ignore last time in .dializer_ignore, but it also now required
the latest TelemetryMetrics, which for the distribution metric type,
moves the buckets option within the reporter options instead of the
metric itself.
```


michalwski (approved):
Thanks for the update!

PR #177, opened 2020-06-22
Add healthcheck endpoint

kmakiela:
* Added healthcheck controller
* Added tests
* Created documentation for this endpoint
* Updated Sparrow


lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,38 @@
...
+
+  def send(conn = %Plug.Conn{}, %{}) do
+    stats = :wpool.stats()
+
+    payload =
+      stats
+      |> Enum.map(&extract_connection_info_from_pool/1)
+      |> Jason.encode!()
+
+    text(conn, payload)
```
rslota: This sends wrong content type (not json) :)

jan-ciesla (approved):
Thanks for the PR, looks good to me.

test/unit/healthcheck_test.exs
```diff
@@ -0,0 +1,94 @@
...
+  import Mox
+
+  setup :verify_on_exit!
+
+  setup do
+    TestHelper.reload_app()
+  end
+
+  test "Successful connection to services" do
+    {_, _, body} = API.get("/healthcheck")
```
michalwski: I think it'd worth to check the status of the response as well. If I understand it correctly this is `200`.

michalwski (commented):
Thanks for implementing the health check. I have 2 points to discuss:
1. Documentation of the `/healthcheck`. I think it's important to say about the endpoint and explain that it's not recommended to do it frequently due to extra work added to the worker processes.
2. Is it possible, does it make sense to add spec of this endpoint to our OpenAPI spec?
3. Open question. Is status 200 a good fit for a situation when all connections are dead? Software like load balancers decides if a service is a health based on the HTTP status code.

test/unit/healthcheck_test.exs
kmakiela: >     3. Open question. Is status 200 a good fit for a situation when all connections are dead? Software like load balancers decides if a service is a health based on the HTTP status code.
I agree, I put 503 status response in that case. However, I'm not completely sure if this "service unavailable" would be about APNS and FCM or the MPush itself

mix.exs
```diff
@@ -31,7 +31,7 @@ defmodule MongoosePush.Mixfile do
   defp deps do
     [
       {:chatterbox, github: "joedevivo/chatterbox", ref: "1f4ce4f", override: true},
-      {:sparrow, github: "esl/sparrow", ref: "b1896ca"},
+      {:sparrow, github: "esl/sparrow", ref: "80a17bd"},
```
NelsonVides: I'm only scared that this might clash with what has just been merged to master when removing :mock, this commit hash is older than that one.

test/support/api.ex
```diff
@@ -30,7 +30,7 @@ defmodule MongoosePush.Support.API do
 
   def get(path) do
     %Response{status_code: status, headers: headers, body: body} =
-      HTTPoison.get!("https://localhost:8443" <> path)
+      HTTPoison.get!("https://localhost:8443" <> path, [], hackney: [:insecure])
```
NelsonVides, line 33: Just out of curiosity, why's that necessary?

README.md
```diff
@@ -490,6 +490,35 @@ If you specify both **alert** and **data**, target device will receive both noti
 * **520** `{"reason" : "unspecified"}` - the unknown error occurred.
 * **500** `{"reason" : reason}` - the server internal error occurred,
   specified by **reason**.
+ 
+### Healthcheck
+
+MongoosePush exposes `/healthcheck` endpoint, from which you can get information about current status of connections in a `JSON` format, grouped by connection pool. An example with 2 pools, one being connected to the service and the other one not would look like this:
```
NelsonVides:
```suggestion
MongoosePush exposes a `/healthcheck` endpoint, from which you can get information about the current status of all connections in a `JSON` format, grouped by connection pool. An example with 2 pools, one being connected to the service and the other one not, would look like this:
```

NelsonVides (commented):
Just an article check and some small questions 🙂

mix.exs
kmakiela: yes, we have to resolve that manually, but it's not a problem

test/support/api.ex
kmakiela: It's necessary because of `bad certificate` problems. We passed this option in `MongoosePush.Support.API.post` and not this one because it wasn't used anywhere in the tests anywhere so far.

test/unit/healthcheck_test.exs
rslota: Maybe this can be used as a guideline: https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html :)

michalwski: That's good reading. Thanks for that @rslota. @kmakiela could you take this document into account?

README.md
```diff
@@ -490,6 +490,35 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+MongoosePush exposes a `/healthcheck` endpoint, from which you can get information about the current status of all connections in a `JSON` format, grouped by connection pool. An example with 2 pools, one being connected to the service and the other one not, would look like this:
+
+```json
+[
+{
+  "pool": "pool_name1",
+  "connection_status":
+    [
+      "connected",
+      "connected"
```
aleklisi: So if I get 100 workers does it mean that this list is 100 "connected" strings?
How about:
```json
"connection_status": {
  "connected": 2,
  "disconnected": 0
}
```

kmakiela: At first I thought we wanted to possibly give more information about the workers but since we stick with the connection status only I agree it's cleaner that way. However, the whole response is going to be restructured to match the draft RFC mentioned earlier

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,115 @@
...
+    }
+  end
+
+  defp get_status(connections) do
+    is_everything_disconnected = Enum.all?(connections, &is_pool_disconnected?/1)
+
+    case is_everything_disconnected do
+      true ->
+        503
+
```
michalwski, line 47: I think this empty line is redundant and make the code less readable.
```suggestion
```

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  defp is_pool_disconnected?(pool_info) do
+    case pool_info[:connection_status][:connected] do
+      0 ->
+        true
+
+      _ ->
+        false
+    end
```
michalwski: How about just this:
```suggestion
    not (pool_info[:connection_status][:connected] > 0)
```

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,115 @@
...
+  defp format_response(status, connections) do
+    {_, pool_infos} =
+      connections
+      |> Enum.map_reduce(%{}, &format_pool_info/2)
+
+    health =
+      case status do
+        503 ->
+          "fail"
+
```
michalwski, line 77:
```suggestion
```

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,115 @@
...
+      details: pool_infos
+    }
+  end
+
+  defp format_pool_info(pool_info, acc) do
+    status =
+      case is_pool_disconnected?(pool_info) do
+        true ->
+          "fail"
+
```
michalwski, line 95:
```suggestion
```

michalwski (commented):
Thanks for making the response compatible with the proposed RFC and adding documentation. I had few additional, minor comments. Other than that it looks good.

kmakiela:
> Thanks for making the response compatible with the proposed RFC and adding documentation. I had few additional, minor comments. Other than that it looks good.

These additional lines are all added and required by `mix.format`.

NelsonVides:
> > Thanks for making the response compatible with the proposed RFC and adding documentation. I had few additional, minor comments. Other than that it looks good.
> 
> These additional lines are all added and required by `mix.format`.

Yeah, often `mix.format` forced me to do unexpected stuff 🤷‍♂️

lib/mongoose_push_web/controllers/healthcheck_controller.ex
leszke: or could it be even `pool_info[:connection_status][:connected] == 0`?

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,119 @@
+defmodule MongoosePushWeb.HealthcheckController do
+  use MongoosePushWeb, :controller
+
+  def send(conn = %Plug.Conn{}, %{}) do
+    stats = :wpool.stats()
+
+    connections =
+      stats
+      |> Enum.map(&extract_connection_info_from_pool/1)
```
leszke: The coding style remark is that it is considered a bad practise to use pipe only once. Please follow [the style guide](https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines) in future.

kmakiela: Good catch, there were more pipes earlier that got removed

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,119 @@
...
+
+  defp extract_connection_info_from_pool(pool) do
+    pool_pid = pool[:supervisor]
+    children = Supervisor.which_children(pool_pid)
+    {_, sup_pid, _, _} = List.keyfind(children, [:wpool_process_sup], 3)
+    workers = Supervisor.which_children(sup_pid)
+
+    {_workers, connection_status} =
+      workers
+      |> Enum.map_reduce(%{connected: 0, disconnected: 0}, fn worker_info, acc ->
```
leszke: And here as well :)

README.md
```diff
@@ -490,6 +490,44 @@ If you specify both **alert** and **data**, target device will receive both noti
 * **520** `{"reason" : "unspecified"}` - the unknown error occurred.
 * **500** `{"reason" : reason}` - the server internal error occurred,
   specified by **reason**.
+ 
+### Healthcheck
+
+MongoosePush exposes a `/healthcheck` endpoint, from which you can get information about the current status of all connections in a `JSON` format, grouped by connection pool. Response structure is described in a following [RFC draft](https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html). An example with 2 pools, one being connected to the service and the other one not, would look like this:
```
NelsonVides:
```suggestion
MongoosePush exposes a `/healthcheck` endpoint, from which you can get information about the current status of all connections in a `JSON` format, grouped by connection pool. The response structure is described in the following [RFC draft](https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html). An example with 2 pools, one being connected to the service and the other one not, would look like this:
```

README.md
```diff
@@ -490,6 +490,44 @@ If you specify both **alert** and **data**, target device will receive both noti
...
+        "time": "2020-07-01T11:58:30.102291Z"
+      }
+    ]
+  },
+  "releaseID": "2.0.2",
+  "status": "pass",
+  "version": "2"
+}
+```
+If all the connections are down the response status is `503` and in all the other cases it's `200`.
```
NelsonVides:
```suggestion
If all the connections are down the response status is `503`; in all the other cases, it's `200`.
```

README.md
NelsonVides: Also, what about updating the link to the newest version maybe?

lib/mongoose_push_web/controllers/healthcheck_controller.ex
```diff
@@ -0,0 +1,114 @@
...
+        200
+    end
+  end
+
+  defp is_pool_disconnected?(pool_info) do
+    pool_info[:connection_status][:connected] == 0
+  end
+
+  # Response formatted to match the draft RFC for healthcheck endpoints, described here:
+  # https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html
```
NelsonVides: Same idea about the link, but only if we're actually compliant to the latest version.

NelsonVides (approved):
Excellent 😄

PR #178, opened 2020-06-22
Pass the logformat variable to function setting the formatter

michalwski:
Previously the `loglevel` was accidentally passed and our tests didn't catch it.


NelsonVides (approved):
Amazing, thanks for that catch! 🎉

PR #179, opened 2020-06-24
update depencencies

leszke:
This petty PR is only about dependencies update. The long-term goal is to make `MongoosePush` fully compatible with `OTP23`.


leszke:
At the beginning I wanted to paste quite a big `cowboy`, `erl` and `iex` versions investigation results, but to not bore our readership with errors and stacktraces, at this moment we just have to stick to the `cowboy 2.7.0` version, as it  needs `OTP22` version, what is beyond the scope of this PR.

michalwski (approved):

PR #181, opened 2020-06-25
Remove obsolete :mock and :meck dependencies

kmakiela:
This PR removes `:meck` and `:mock` from the dependency list.
Both `:meck` and `:mock` are not needed anymore. All the tests that used these libraries were already rewritten (most of them use `:mox` now).


NelsonVides:
So cool it would be so simple to get rid of Mock 😄 
But what about the `mix.lock` file, doesn't that one need to be cleaned out as well?

leszke (approved):

michalwski:
@kmakiela I second @NelsonVides questions. Looks like mix.lock needs an update.

kmakiela:
Further investigation shows that `:meck` is still loaded as a Sparrow dependency (and Sparrow only) so this PR is temporarily blocked by:
https://github.com/esl/sparrow/pull/47

NelsonVides:
@kmakiela https://github.com/esl/sparrow/pull/47 was merged, we can move forward with this one 🙂

NelsonVides (approved):
Neat 🎉

PR #182, opened 2020-06-25
move from rslota/beam-builder to mongooseim/beam-builder

leszke:
As per PR description, we are moving to esl fork of `beam-builder`, which also means we are moving to `mongooseim/beam-builder` repository on Docker Hub, which I populated with proper images today.


michalwski (approved):
Thanks!

PR #184, opened 2020-07-03
Deploy documentation to GH Pages from CCI

zofpolkowska:
Add a CCI job which builds new documentation for every new tag and updates the latest version on every commit to master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,12 +299,6 @@ jobs:
               sudo apt-get update && \
               sudo apt-get install -y esl-erlang=1:21.0 && \
               sudo apt-get install -y elixir=1.8.0-1
-              wget https://github.com/jwilder/dockerize/releases/download/v0.6.1/dockerize-linux-amd64-v0.6.1.tar.gz && \
-              sudo chmod +x dockerize-linux-amd64-v0.6.1.tar.gz && \
-              sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-v0.6.1.tar.gz && \
-              sudo chmod +x /usr/local/bin/dockerize && \
-              sudo rm dockerize-linux-amd64-v0.6.1.tar.gz
-
       - restore_cache:
           keys:
               - mix-cache-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -317,8 +317,8 @@ jobs:
       - run:
           name: Configure Git
           command: |
-              git config --global user.email "builds@circleci.com"
-              git config --global user.name "CircleCI"
+              git config --global user.email "mongoosepushbot@erlang-solutions.com"
+              git config --global user.name "mongoosepushbot"
       - add_ssh_keys:
           fingerprints:
             - "${GH_DEPLOY_KEY}"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -286,6 +286,53 @@ jobs:
                docker push $DOCKERHUB_REPOSITORY/$DOCKER_IMAGE:latest
              fi
 
+  gh-pages-deploy:
+    machine:
+        image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - run:
+          name: Install elixir
+          command: |
+              wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb && \
+              sudo dpkg -i erlang-solutions_1.0_all.deb && \
+              sudo apt-get update && \
+              sudo apt-get install -y esl-erlang=1:21.0 && \
+              sudo apt-get install -y elixir=1.8.0-1
+              wget https://github.com/jwilder/dockerize/releases/download/v0.6.1/dockerize-linux-amd64-v0.6.1.tar.gz && \
+              sudo chmod +x dockerize-linux-amd64-v0.6.1.tar.gz && \
+              sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-v0.6.1.tar.gz && \
+              sudo chmod +x /usr/local/bin/dockerize && \
+              sudo rm dockerize-linux-amd64-v0.6.1.tar.gz
+
+      - restore_cache:
+          keys:
+              - mix-cache-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}
+      - restore_cache:
+          keys:
+              - certs-cache-{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-{{ .Revision }}
+      - run:
+          name: Compile mix deps
+          command: mix do deps.get, deps.compile
+      - run:
+          name: Configure Git
+          command: |
+              git config --global user.email "builds@circleci.com"
+              git config --global user.name "CircleCI"
+      - add_ssh_keys:
+          fingerprints:
+            - "${GH_DEPLOY_KEY}"
+      - run:
+          name: Generate docs
+          command: |
+              set -x
+              if [ -n "$CIRCLE_TAG" ]; then
+                  mix gh_pages_docs $CIRCLE_TAG
+              elif [ "$CIRCLE_BRANCH" == "master" ]; then
+                  mix gh_pages_docs latest
+              fi
+
+
 workflows:
   version: 2
   build_and_test:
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,82 @@ defmodule Mix.Tasks.GhPagesDocs do
...
+          update_versions_js()
         end
 
+        update_index_html()
+
+        Mix.shell().cmd("git add assets/js/versions.js")
+        Mix.shell().cmd("git add index.html")
+        Mix.shell().cmd("git commit -m \"Add content for #{version}\"")
+        Mix.shell().cmd("rm -rf doc")
+        Mix.shell().cmd("git show ")
```
NelsonVides: Why do we need `git show` here?

NelsonVides (commented):
Thanks a lot for this feat! Seems awesome, I'm waiting to see it in action!
I just have one question on one of the lines 🙂

lib/mix/tasks/gh_pages_docs.ex
zofpolkowska: We don't 🙃

.circleci/config.yml
```diff
@@ -286,6 +286,47 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install elixir
+          command: |
+              wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb && \
+              sudo dpkg -i erlang-solutions_1.0_all.deb && \
+              sudo apt-get update && \
+              sudo apt-get install -y esl-erlang=1:21.0 && \
+              sudo apt-get install -y elixir=1.8.0-1
```
michalwski, line 301: Does it make a difference in the UI or structure of the generated files if newer Elixir version is used?

.circleci/config.yml
```diff
@@ -286,6 +286,47 @@ jobs:
...
+      - add_ssh_keys:
+          fingerprints:
+            - "${GH_DEPLOY_KEY}"
+      - run:
+          name: Generate docs
+          command: |
+              set -x
+              if [ -n "$CIRCLE_TAG" ]; then
+                  mix gh_pages_docs $CIRCLE_TAG
+              elif [ "$CIRCLE_BRANCH" == "cci-doc" ]; then
```
michalwski: I think we have a leftover here.
```suggestion
              elif [ "$CIRCLE_BRANCH" == "master" ]; then
```

.circleci/config.yml
```diff
@@ -333,3 +374,12 @@ workflows:
...
           filters: *all_tags
+
+      - gh-pages-deploy:
+          requires:
+            - pre-build
+          filters:
+            tags:
+              only: /^\d+\.\d+\.\d+([a-z0-9\-\+\.])*/
+            branches:
+              only: cci-doc
```
michalwski: Another change to the `master` branch. I wonder if we want to do it in the PR or after it's merged.

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,82 @@ defmodule Mix.Tasks.GhPagesDocs do
...
-            Mix.shell().cmd("ls")
-            Mix.shell().cmd("git add #{version}/*")
-            Mix.shell().cmd("git commit -m \"Update #{version}\"")
-            Mix.shell().cmd("rm -rf doc")
-            Mix.shell().cmd("git push origin gh-pages")
-
-          _ ->
-            Mix.raise("Cannot create the #{version} directory")
+  def run([version]) do
+    unless File.read!("assets/js/versions.js") |> String.contains?(version) do
```
michalwski: In one of the other PR's review, I noticed that we try to avoid the pipeline operator in case it's used only once. Do we want to keep the convention here as well?

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,82 @@ defmodule Mix.Tasks.GhPagesDocs do
...
+
+    version =
+      case version do
+        "latest" -> Project.config()[:version]
+        tag -> "v#{tag}"
+      end
+
+    Mix.Task.run("docs")
+    File.cd!("doc")
+    Mix.shell().cmd("git stash")
```
michalwski: At first, I thought that after `git stash` we loose the just generated docs. Then, I realised that the `doc` dir is added to `.gitignore` so `git stash` doesn't touch it. Hence, we can access it when we switch to the `gh-pages` branch. I think it'd clearer for other colleagues reading this script if a comment similar to this is added to the code.

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,82 @@ defmodule Mix.Tasks.GhPagesDocs do
...
   end
+
+  defp update_versions_js() do
+    current = Project.config()[:version]
+    {output, 0} = System.cmd("git", ["tag"], [])
+
+    versions =
+      output
+      |> String.split("\n")
+      |> Enum.drop(7)
```
michalwski: I'm afraid this is a bit error-prone. If for some reasons a tag which today has number 4 is removed, then index 7 contains some other tag, which we probably don't want to remove. Is it possible to explicitly say which tags we are skipping?

michalwski (commented):
Thanks for the changes. It's good to see that the process can be automated on CI! I had a few comments and questions on the code.

.circleci/config.yml
zofpolkowska: when the PR is good to merge, then I'll revert this temp change for testing

.circleci/config.yml
zofpolkowska: when the PR is good to merge, then I'll revert this temp change for testing

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,93 @@ defmodule Mix.Tasks.GhPagesDocs do
...
+        "latest" -> Project.config()[:version]
+        tag -> "v#{tag}"
+      end
+
+    Mix.Task.run("docs")
+    File.cd!("doc")
+    Mix.shell().cmd("git stash")
+
+    # ignore all the changes, the only changes we want to track are in the
+    # doc directory which is listed .gitignore
```
NelsonVides:
```suggestion
    # doc directory which is listed in .gitignore
```

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,93 @@ defmodule Mix.Tasks.GhPagesDocs do
...
+      | String.split(output, "\n") --
+          [
+            "0.1.0",
+            "0.10.0",
+            "0.9.0",
+            "1.0.0",
+            "1.0.3",
+            "1.0.4",
+            "1.0.5"
+          ]
```
NelsonVides, line 72: Maybe we can simply put a comment explaining that we're just not generating docs for these tags, just so it's faster to understand in the future

NelsonVides (commented):
So, very small comments, it's basically ready to merge, I think you can push with the master tags already 😄

lib/mix/tasks/gh_pages_docs.ex
zofpolkowska: Sure

.circleci/config.yml
NelsonVides: Some dependencies seem to break with Elixir 1.8, and nevertheless these are the versions used in integration tests, so I think we could consider upgrading versions as part of a different ticket 🙂

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -10,36 +10,93 @@ defmodule Mix.Tasks.GhPagesDocs do
...
+    File.cd!("doc")
+    Mix.shell().cmd("git stash")
+
+    # ignore all the changes, the only changes we want to track are in the
+    # doc directory which is listed in .gitignore
+
+    Mix.shell().cmd("git checkout gh-pages")
+
+    case File.mkdir("../#{version}") do
+      result when result in [:ok, {:error, :eexist}] ->
```
aleklisi, line 34: Should it be `:eexist` or `:exist`?

aleklisi (commented):
Would it make sense for each `Mix.shell().cmd("costam costam")` to match the result and check if it was successful?
```elixir
0 = Mix.shell().cmd("costam costam")
```

Other than that I do not see any other major issues :)

lib/mix/tasks/gh_pages_docs.ex
zofpolkowska: https://hexdocs.pm/elixir/File.html#mkdir/1 `eexist` :)

aleklisi (approved):

NelsonVides (approved):
🎉

PR #185, opened 2020-07-13
Update config guide with logformat var

kmakiela:
It seems like when this logging change was introduced the documentation was not updated accordingly

### DIFF
--- a/guides/configuration.md
+++ b/guides/configuration.md
@@ -112,6 +112,7 @@ Environment variables to configure a production release.
 
 #### General settings:
 * `PUSH_LOGLEVEL` - `debug`/`info`/`warn`/`error` - Log level of the application. `info` is the default one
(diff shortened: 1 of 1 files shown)


guides/configuration.md
```diff
@@ -112,6 +112,7 @@ Environment variables to configure a production release.
 
 #### General settings:
 * `PUSH_LOGLEVEL` - `debug`/`info`/`warn`/`error` - Log level of the application. `info` is the default one
+* `PUSH_LOGFORMAT` - `logfmt`/`json` - Log format of the application. Defaults to `logfmt` for `dev` and `test` environments and to `json` for `prod` environment.
```
NelsonVides:
```suggestion
* `PUSH_LOGFORMAT` - `logfmt`/`json` - Log format of the application. Defaults to `logfmt` for the `dev` and `test` environments, and to `json` for the `prod` environment.
```
As we're mentioning specific environments, it needs a definite article. Wouldn't need one if it was an indefinite plural, like "for development environments", and not "for the development environments", but in this case the envs have specific names 🤷‍♂️

NelsonVides (approved):

PR #198, opened 2021-11-12
APNS Silent notification fix (iOS 13+)

### LINKED ISSUE #105: Push notifications with iOS 13 don't work
Hi all,

We were testing our implementation of mongooseIM (version 3.4.0) with mongoosePush(1.0.5) with 2 iPhone, one with iOS 12 and one with iOS 13. We use a simple XMPP client.
On iOS 12 push notifications work ok, but on iOS 13 they don't work at all: we don't receive push notifications.
We are checking if it is a client issue, meanwhile, could you check if mongoosePush is iOS 13 compatible?

Thank you

4nirmall:
APNS Silent notification fix

1. Added "apns-push-type" header with value "background" as this is mandatory for iOS 13+
2. This fixes #105

### DIFF
--- a/lib/mongoose_push/service/apns.ex
+++ b/lib/mongoose_push/service/apns.ex
@@ -13,17 +13,27 @@ defmodule MongoosePush.Service.APNS do
   alias MongoosePush.Service.APNS.ErrorHandler
 
   @priority_mapping %{normal: "5", high: "10"}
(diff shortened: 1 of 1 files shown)


gustawlippa:
Thanks for the contribution!
This should indeed help with issues on newer iOS versions, and wearOS:
obraz

PR #203, opened 2022-09-14
Upgrade dependencies

kamilwaz:
Updates Phoenix and other dependencies to latest versions. Alongside, Erlang was updated to 25.0 and Elixir to 1.13.


kamilwaz:
Erlang cannot be upgraded to 25.x because `distillery` doesn't support this version yet. The change that blocks it: https://github.com/erlang/otp/commit/5ea1d07c17447ff114879b292d4b9d461c191ccd.

Dockerfile
```diff
@@ -23,7 +30,7 @@ RUN mix do certs.dev, distillery.release
 RUN tar -czf mongoose_push.tar.gz -C _build/prod/rel/mongoose_push .
 
 
-FROM debian:stretch-slim
+FROM ubuntu:20.04
```
michalwski: This is the final docker image inside which the built MongoosePush will run. We used `debian:stretch-slim` to make the final image as small as possible. Do we have something similar to `ubuntu:20.04` ?

test/unit/test_helper.exs
```diff
@@ -1,2 +1,4 @@
+Code.put_compiler_option(:warnings_as_errors, true)
```
michalwski, line 1: 👍

michalwski (commented):
Looks good, I had some comments/questions.

Dockerfile
kamilwaz: Unfortunately, not. I changed the final image back to `debian`.

michalwski (approved):
Ok, approving it. Go ahead and merge.

michalwski:
> Erlang cannot be upgraded to 25.x because distillery doesn't support this version yet. The change that blocks it: https://github.com/erlang/otp/commit/5ea1d07c17447ff114879b292d4b9d461c191ccd.

In fact, we could move away from `distillery` to the built-in release mechanism. The built-in got improved in Elixir after this project was actively developed so probably there's no reason to use the `distillery` anymore.

PR #204, opened 2022-09-20
Migrate to mix release

kamilwaz:
Migrate releases from `distillery` to `mix release`.


mix.exs
```diff
@@ -39,9 +40,7 @@ defmodule MongoosePush.Mixfile do
       {:plug_cowboy, "~> 2.2"},
       {:jason, "~> 1.4"},
       {:poison, "~> 3.0", override: true},
-      {:distillery, "~> 2.0", override: true},
       {:confex, "~> 3.2", override: true},
-      {:mix_docker, "~> 0.5"},
```
michalwski, line 44: Why the `mix_docker` was removed?

michalwski (commented):
I had only one question.

michalwski:
One more thing. Could you point to a CircleCi log showing that MongoosPush starts in docker and we can talk to it like we did before?

mix.exs
kamilwaz: I removed it because it wasn't used anywhere at all. Additionally, this library is no longer maintained.

kamilwaz:
> One more thing. Could you point to a CircleCi log showing that MongoosPush starts in docker and we can talk to it like we did before?

Here: https://app.circleci.com/pipelines/github/esl/MongoosePush/1191/workflows/9af95ef1-db85-4343-829f-c25532340608/jobs/9482?invite=true#step-111-2. 

`mix test.env.wait` waits for all components to be up and running, including MongoosePush (https/8443).

michalwski (approved):
Approving

PR #214, opened 2024-03-27
Fix and refactor job for publishing documentation

pawlooss1:
This PR:
* Updates SSH configuration for the job. I also generated a new key with write permissions and added it to the CircleCI project.
* Updates CI badges.
* Fixes bugs in `gh_pages_docs.ex`.
* Introduces a new way of publishing the "latest" documentation. Now the mix project version between releases is set with the suffix `-dev`. This way we treat it very similarly to the releases while having consistent drop-down list with versions at our documentation site.

I ran the new job on the branch to ensure that it succeeds. They can be viewed [here](https://app.circleci.com/pipelines/github/esl/MongoosePush/1281/workflows/d990551f-1bea-4f09-9696-311ec903b1c5/jobs/10082) for the release `2.1.1` and [here](https://app.circleci.com/pipelines/github/esl/MongoosePush/1290/workflows/56756046-1880-4726-b048-837b09c8a050/jobs/10142) for the latest master.
Generated documentation is available [here](https://esl.github.io/MongoosePush/v2.2.0-dev/readme.html).

### DIFF
--- a/lib/mix/tasks/gh_pages_docs.ex
+++ b/lib/mix/tasks/gh_pages_docs.ex
@@ -9,16 +9,12 @@ defmodule Mix.Tasks.GhPagesDocs do
 
   @shortdoc "Update a documentation version on GH Pages."
 
-  @spec run(term) :: :ok
+  @spec run([String.t()]) :: :ok
   def run([version]) do
-    unless String.contains?(File.read!("assets/js/versions.js"), version) do
-      update_versions_js()
-    end
-
     version =
       case version do
         "latest" -> "v#{Project.config()[:version]}"
-        tag -> "v#{tag}"
+        tag -> prefix_tag(tag)
       end
 
     Mix.Task.run("docs")
@@ -37,7 +33,7 @@ defmodule Mix.Tasks.GhPagesDocs do
         0 = Mix.shell().cmd("git add #{version}/*")
 
         unless String.contains?(File.read!("assets/js/versions.js"), version) do
-          update_versions_js()
+          update_versions_js(version)
         end
 
         update_index_html()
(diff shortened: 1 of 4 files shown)


lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -1,91 +1,88 @@
...
-
     version =
       case version do
-        "latest" -> "v#{Project.config()[:version]}"
-        tag -> "v#{tag}"
+        "latest" -> prefix_tag(Mix.Project.config()[:version])
+        tag -> prefix_tag(tag)
       end
 
+    # firstly we need to update version.js for the mix docs task
```
chrzaszcz:
```suggestion
    # firstly we need to update versions.js for the mix docs task
```

lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -13,35 +11,32 @@ defmodule Mix.Tasks.GhPagesDocs do
...
+
     Mix.Task.run("docs")
-    File.cd!("doc")
+
     0 = Mix.shell().cmd("git stash")
+    0 = Mix.shell().cmd("git checkout gh-pages")
 
-    # ignore all the changes, the only changes we want to track are in the
-    # doc directory which is listed in .gitignore
+    # secondly we do it again to avoid git conflicts from git stash pop
```
chrzaszcz: Could you explain where the `git stash pop` is executed?

chrzaszcz (commented):
Looks good. I added a comment and a question.

lib/mix/tasks/gh_pages_docs.ex
pawlooss1: I think I took a mental shortcut that was too big. What I meant is that we could use the stash entry with `versions.js` that was created before the checkout but it would cause a git conflict when executing `git stash pop`. I decided to invoke `update_versions_js/1` once again as a simpler way instead of resolving such conflict.
I'll try to make it more clear in this comment.

chrzaszcz (approved):
👌

PR #216, opened 2024-04-10
Fix config resolution

### LINKED ISSUE #206: Crash at startup with token.p8 - bad_auth
I am trying to start MongoosePush ver 2.1.0 from the command line:
`_build/prod/rel/mongoose_push/bin/mongoose_push start`

It would crash with the following errors:
```
{
   "application":"mongoose_push",
   "at":"Elixir.MongoosePush.Service.APNS.Supervisor.convert_token_pool_to_sparrow/2:150",
   "erl_level":"error",
   "file":"lib/mongoose_push/service/apns/supervisor.ex",
   "key":null,
   "mode":"dev",
   "p8_file":"/home/developer/MongoosePush-master/_build/prod/rel/mongoose_push/lib/mongoose_push-2.1.0/priv/apns/token.p8",
   "pid":"#PID<0.1801.0>",
   "reason":"bad_auth",
   "result":"error",
   "severity":"error",
   "team":null,
   "text":"Required configuration missing",
   "what":"configuration",
   "when":"2022-10-02T19:02:12.722",
   "domain[0]":"elixir"
}
```

```
{
  "application": "mongoose_push",
  "at": "Elixir.MongoosePush.Service.APNS.Supervisor.convert_token_pool_to_sparrow/2:150",
  "erl_level": "error",
  "file": "lib/mongoose_push/service/apns/supervisor.ex",
  "key": null,
  "mode": "prod",
  "p8_file": "/home/developer/MongoosePush-master/_build/prod/rel/mongoose_push/lib/mongoose_push-2.1.0/priv/apns/token.p8",
  "pid": "#PID<0.1801.0>",
  "reason": "bad_auth",
  "result": "error",
  "severity": "error",
  "team": null,
  "text": "Required configuration missing",
  "what": "configuration",
  "when": "2022-10-02T19:02:12.722",
  "domain[0]": "elixir"
}
```
I have copied the files from Apple into the `/home/developer/MongoosePush-master/_build/prod/rel/mongoose_push/lib/mongoose_push-2.1.0/priv` directories.

Does anyone have  any idea or solution about these?

It says "Required configuration missing" but it doesn't seem to be using any files under the `config` directory.

Many thanks in advance.

pawlooss1:
This PR changes resolution of configuration environmental variables.
Previously the resolver would iterate over all configuration keys and call `Confex.fetch_env!(app, key)` to resolve all values. The issue with it was that if some keys were already set by TOML resolver they would still be seen as `{:system, type, var_name, default}` which led to overriding TOML config with defaults.
Also, it changes default values of `PUSH_*_ENABLED` to false to avoid implicit creation of an incorrect pool.
At last it fixes a clash where after defining some service pools in TOML file, entries from `prod.exs` would be appended to them due to the behaviour of `Config.Reader.merge/2`. If someone defines pools in TOML file then no other pools should be created.

### DIFF
--- a/lib/mongoose_push/config/provider/toml.ex
+++ b/lib/mongoose_push/config/provider/toml.ex
@@ -54,7 +54,8 @@ defmodule MongoosePush.Config.Provider.Toml do
       case toml[:general][:logging][:level] do
         "debug" -> :debug
         "info" -> :info
-        "warn" -> :warn
+        "warn" -> :warning
+        "warning" -> :warning
         "error" -> :error
         nil -> sysconfig[:logging][:level]
         invalid -> raise "Invalid loglevel: #{invalid}!"
--- a/lib/mongoose_push/config/provider/toml.ex
+++ b/lib/mongoose_push/config/provider/toml.ex
@@ -52,8 +52,8 @@ defmodule MongoosePush.Config.Provider.Toml do
   end
 
   defp maybe_erase_confex_services(config, updated_sysconfig) do
-    # If some pools are defined in TOML config we want to remove the default ones explicitely
-    # because Config.Reader.merge/2 would keep them both and the default would still be created
+    # If some pools are defined in TOML config, we want to remove the default ones explicitly
+    # because Config.Reader.merge/2 would keep them both, and the default would still be created
     config
     |> maybe_erase_default_service(:fcm, updated_sysconfig[:fcm_enabled])
     |> maybe_erase_default_service(:apns, updated_sysconfig[:apns_enabled])
--- a/lib/mongoose_push/config/provider/toml.ex
+++ b/lib/mongoose_push/config/provider/toml.ex
@@ -17,6 +17,8 @@ defmodule MongoosePush.Config.Provider.Toml do
       current_sysconfig = Application.get_all_env(@app)
       updated_sysconfig = update_sysconfig(current_sysconfig, toml_config)
 
+      config = maybe_erase_confex_services(config, updated_sysconfig)
+
       Config.Reader.merge(
         config,
         [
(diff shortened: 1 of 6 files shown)


lib/mongoose_push/config/provider/toml.ex
```diff
@@ -49,12 +51,21 @@ defmodule MongoosePush.Config.Provider.Toml do
     |> update_service(toml_config, :apns)
   end
 
+  defp maybe_erase_confex_services(config, updated_sysconfig) do
+    # If some pools are defined in TOML config we want to remove the default ones explicitely
+    # because Config.Reader.merge/2 would keep them both and the default would still be created
```
chrzaszcz:
```suggestion
    # If some pools are defined in TOML config, we want to remove the default ones explicitly
    # because Config.Reader.merge/2 would keep them both, and the default would still be created
```

lib/mongoose_push/config/provider/toml.ex
```diff
@@ -17,11 +17,13 @@ defmodule MongoosePush.Config.Provider.Toml do
       current_sysconfig = Application.get_all_env(@app)
       updated_sysconfig = update_sysconfig(current_sysconfig, toml_config)
 
+      config = maybe_erase_confex_services(config, updated_sysconfig)
```
chrzaszcz, line 20: How do we know that the Confex provider is applied before the TOML one?

chrzaszcz (commented):
Looks good in general. I just added a question.

lib/mongoose_push/config/provider/toml.ex
pawlooss1: The order is defined in `mix.exs` [here](https://github.com/esl/MongoosePush/blob/master/mix.exs#L201). Actually it's the other way - TOML is applied before Confex. Here we're erasing Confex entries before they're even resolved. For instance, in [prod.exs](https://github.com/esl/MongoosePush/blob/master/config/prod.exs#L48) we have a `default` fcm pool which is composed of Confex entries in format `{:system, type, var_name, default}`. We want to ensure that this whole pool is erased if other pools are defined by TOML config. Same goes for apns.

chrzaszcz: Sounds good. I knew the order from `mix.exs`, but I incorrectly assumed that we are removing the resolved pools.

chrzaszcz (approved):
Looks good 👍

PR #220, opened 2025-05-13
Add migration guides

pawlooss1:
This PR add "Migration Guides" section to the hex documentation. Also, it attaches previous migration guide that was in the repository but not published and adds a migration guide from 2.1.1 to 2.2.2.
 To check how does it look run `mix docs` and open `doc/index.html` file.


chrzaszcz (approved):
Looks good 👍

PR #221, opened 2025-05-20
Fix gh pages deploy

pawlooss1:
This PR is a follow-up to https://github.com/esl/MongoosePush/pull/220. It fixes `gh-pages-deploy` job which runs only on master branch. The fix is similar to the one before for `integration-tests`, i.e. install Erlang and Elixir from rabbitmq repository instead of esl.


chrzaszcz (approved):
Looks good 👍

PR #222, opened 2025-05-22
Update dependencies

pawlooss1:
This PR updates the dependencies and also Erlang and Elixir used in CI. In CI we test two most recent Erlang and Elixir versions. Only `httpoison` required code changes (in test helpers only) as it changed the way of creating ssl options.


chrzaszcz (approved):
Looks good 👍

PR #224, opened 2025-06-10
Dry run docs

pawlooss1:
This PR add dry-run of the `gh-pages-deploy` CI job. It allows to run and debug it on any branch which should spare us follow-up fixes after it fails on master branch.
Additionally it is fixed by bumping ubuntu version of the executor.


jacekwegr (approved):
Looks good! Good idea with the dry run 👍

PR #225, opened 2025-06-18
Fix version sidebar in docs

pawlooss1:
New version of `ex_docs` doesn't support `javascript_config_path` option anymore and expects a file called `docs_config.js` present in a directory with documentation.
For comparison see:
- documentation generated with older `ex_docs` - https://github.com/esl/MongoosePush/blob/gh-pages/v2.1.1/readme.html#L19
- documentation generated with newer `ex_docs` - https://github.com/esl/MongoosePush/blob/gh-pages/v2.2.0/readme.html#L16

The mix task is modified to put such file in every documentation directory so the newer docs can use it (it is ignored for older versions).

Also `assets/js/versions.js` is removed from the repository as it's no longer needed.


lib/mix/tasks/gh_pages_docs.ex
```diff
@@ -16,9 +16,6 @@ defmodule Mix.Tasks.GhPagesDocs do
         tag -> {prefix_tag(tag), false}
       end
 
-    # Firstly we need to update versions.js for the mix docs task
-    update_versions_js(version)
```
jacekwegr, line 20: Later in the file, there are lines:
```exs
# Secondly we do it again to recreate it after checkout
# It is a simpler way than using git stash pop and resolving conflicts
update_versions_js(version)
```
Could you also update this comment to reflect what's going on now?

jacekwegr (commented):
Looks good! I just added one minor comment.

lib/mix/tasks/gh_pages_docs.ex
pawlooss1: Good catch 👍 Now the "Secondly" doesn't make any sense 😅

jacekwegr (approved):
Looks good! 👍

PR #227, opened 2026-01-21
Build multiplatform docker image

pawlooss1:
This PR introduces building multiplatform docker images with `buildx`. Also, it fixes `integration-tests` and `gh-pages-deploy` jobs which started to fail due to debian package issues.
Details in the commit messages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -298,8 +298,6 @@ jobs:
   gh-pages-deploy:
     docker:
       - image: cimg/elixir:1.18.3-erlang-27.2.2
-        environment:
-          MIX_ENV: dev
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -283,17 +296,12 @@ jobs:
              fi
 
   gh-pages-deploy:
-    machine:
-        image: ubuntu-2404:current
+    docker:
+      - image: cimg/elixir:1.18.3-erlang-27.2.2
+        environment:
+          MIX_ENV: dev
```
kamilwaz: I believe this is not needed since the default environment is dev.

kamilwaz (commented):
Looks good :+1: Just left one small comment.

.circleci/config.yml
pawlooss1: Thanks, I've cleaned it up.

kamilwaz (approved):

PR #230, opened 2026-02-18
Fix failing intergration tests

kamilwaz:
Integration tests are failing with the following error:

```
E: Version '1:27.3.4.6-1rmq1ppa1~ubuntu24.04.1' for 'erlang' was not found
```

This PR relaxes the version matches, so the latest packages are installed, which should always come from the PPA:

```bash
$ docker run -it --rm ubuntu:24.04
apt update
apt install -y software-properties-common
apt update
apt policy erlang
```
```
erlang:
  Installed: (none)
  Candidate: 1:27.3.4.7-1rmq1ppa1~ubuntu24.04.1
  Version table:
     1:27.3.4.7-1rmq1ppa1~ubuntu24.04.1 500
        500 https://ppa.launchpadcontent.net/rabbitmq/rabbitmq-erlang/ubuntu noble/main arm64 Packages
     1:25.3.2.8+dfsg-1ubuntu4.6 500
        500 http://ports.ubuntu.com/ubuntu-ports noble-updates/universe arm64 Packages
        500 http://ports.ubuntu.com/ubuntu-ports noble-security/universe arm64 Packages
     1:25.3.2.8+dfsg-1ubuntu4 500
        500 http://ports.ubuntu.com/ubuntu-ports noble/universe arm64 Packages
```
\
This PR also simplifies the CircleCI configuration and replaces `docker-compose` with `docker compose`.

MIM-2578

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ run: &DOCKER_TAGS
         echo "export DOCKER_IMAGE=mongoose-push" >> $BASH_ENV
 
 executors:
-  elixir-builder:
+  elixir-builder: &elixir-builder
     parameters:
       erlang_version:
         type: string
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -248,8 +248,8 @@ jobs:
           command: |
               sudo add-apt-repository ppa:rabbitmq/rabbitmq-erlang -y
               sudo apt update
-              sudo apt-get install -y erlang=1:27.3.4.6-1rmq1ppa1~ubuntu24.04.1
-              sudo apt-get install -y elixir=1.17.3-1rmq1ppa1~ubuntu24.04.1
+              sudo apt-get install -y erlang
```
DenysGonchar: we should better use docker executors instead of VM

kamilwaz: After switching to the docker executor, the tests are failing because we're running docker inside docker. For now, I'll leave it as it is (we'll have ot review the whole thing at some point anyway).

.circleci/config.yml
```diff
@@ -35,26 +35,25 @@ executors:
...
       - image: hexpm/elixir:<< parameters.elixir_version >>-erlang-<< parameters.erlang_version >>-ubuntu-noble-20250415.1
-      - image: erlangsolutions/fcm-mock-server
-      - image: erlangsolutions/apns-mock-server
     working_directory: ~/app
     environment:
         MIX_ENV: << parameters.env >>
+  elixir-env:
+    <<: *elixir-builder
+    docker:
+      - image: hexpm/elixir:<< parameters.elixir_version >>-erlang-<< parameters.erlang_version >>-ubuntu-noble-20250415.1
```
DenysGonchar: please use prebuilt CircleCI images:
https://circleci.com/developer/images/image/cimg/elixir

DenysGonchar: see https://circleci.com/docs/guides/execution-managed/using-docker/#caching-docker-images for more details

.circleci/config.yml
DenysGonchar: the idea is that we don't use docker compose in this case:
https://circleci.com/docs/guides/execution-managed/using-docker/#using-multiple-docker-images

but taking into account the following changes, you have figured this out by yourself :).

DenysGonchar (approved):

